PR #2, opened 2018-10-27
Move `has_issuance` and `is_pegin` fields from OutPoint to TxIn

apoelstra:
To more closely match Elements, and to give users a way to refer to an outpoint independently of its use in an input.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ before_install:
   - sudo apt-get install -y binutils-dev libunwind8-dev
 
 script:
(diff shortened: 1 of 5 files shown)


apoelstra:
cc @jonasnick

cc @RCasatta -- can you take a look at this and see if it breaks anything for GA uses? It bumps the major version number so it won't affect you until you upgrade, but I don't want to take the project in a direction that breaks you.

RCasatta:
I checked the impact on our side and it's minimal

thanks for the heads up!

PR #5, opened 2018-10-30
add `PeginData` to complement `PegoutData`

apoelstra:


apoelstra:
@RCasatta another breaking change .. I think this won't affect you, it only breaks some new methods that Jonas and I added last night, so presumably y'all aren't using them :). Just a heads up.

RCasatta:
cool stuff, this will help with the explorer

PR #6, opened 2018-11-07
Export AssetIssuance

shesek:
Because `elements::transaction` is private, the `AssetIssuance` struct cannot be accessed without this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,6 @@ pub mod confidential;
 mod transaction;
 
 // export everything at the top level so it can be used as `elements::Transaction` etc.
-pub use transaction::{OutPoint, PeginData, PegoutData, TxIn, TxOut, TxInWitness, TxOutWitness, Transaction};
+pub use transaction::{OutPoint, PeginData, PegoutData, TxIn, TxOut, TxInWitness, TxOutWitness, Transaction, AssetIssuance};
 pub use block::{BlockHeader, Block, Proof};
 


jonasnick:
@apoelstra any idea why travis is failing on rust stable?
```
error: failed to run custom build command for `honggfuzz v0.5.30`
process didn't exit successfully: `/home/travis/build/ElementsProject/rust-elements/fuzz/hfuzz_target/release/build/honggfuzz-d658ffe7a4e71193/build-script-build` (exit code: 1)
--- stderr
honggfuzz dependency (0.5.30) and build command (0.5.31) versions do not match
warning: build failed, waiting for other jobs to finish...
error: build failed
The command "if [ "$(rustup show | grep default | grep stable)" != "" ]; then cd fuzz && cargo test --verbose && ./travis-fuzz.sh; fi" exited with 101.
```

apoelstra:
Rebased. Should be good now.

apoelstra:
Oh, hm, my rebasing lost @shesek signature ... maybe you want to sign the commit I just pushed?

shesek:
I made the PR using Github's web interface, it should not have been signed. Was it? Quite weird if so...

apoelstra:
Interesting. Github has now hidden the commit from me so I can't tell, but it said "Verified" on the UI and when I clicked on that it showed me your gpg fingerprint. Probably if I'd read closer I would've seen that it didn't actually mean a gpg signature.

In future I'll be sure to verify signatures locally then :P.

@jonasnick can you ack this one?

PR #8, opened 2018-11-08
Add unit test for asset issuance deserialization

apoelstra:


jonasnick:
how about adding Cargo.lock to .gitignore

src/transaction.rs
```diff
@@ -1112,5 +1111,337 @@ mod tests {
...
+                        0x8b, 0xea, 0x94, 0x99, 0x7d, 0xce, 0x4a, 0xe8,
+                        0x5b, 0x3d, 0x95, 0xa2, 0x07, 0x00, 0x38, 0x4f,
+                        0x0b, 0x8c, 0x1f, 0xe9, 0x95, 0x18, 0x06, 0x38
+                    ],
+                ),
+                inflation_keys: confidential::Value::Null,
+            }
+        );
+        assert_eq!(
+            tx.input[0].asset_entropy().unwrap().to_string(),
```
jonasnick: travis complains that rustc doesn't know `asset_entropy()`

apoelstra: Oh, whoops, I deleted that function (which is incomplete/blocked on bitcoin_hashes exposing midstates) but forgot to remove from the test.

PR #9, opened 2018-11-08
update rust-bitcoin to 0.15; update major version to 0.4

apoelstra:


apoelstra:
#6 should go in first since that's a non-breaking change

PR #11, opened 2018-11-25
transaction: fix `TxOut::is_pegout()` to fail properly on certain invalid inputs

apoelstra:


src/transaction.rs
```diff
@@ -1440,5 +1440,55 @@ mod tests {
...
+            edededededededede9edededede00000000000001000000000000050000ff0000\
+            000000000000000000ff000000000000000000000000200000000000011c00000\
+            000d3d3d3d3d3d3d3d3d3d3d3d3d3d3d3d3d3d3f3d3d3d3d3d3d3d3d3d3d3d3d3\
+            d3d3d3d3d3d3\
+        ");
+
+        assert!(!output.is_null_data());
+        assert!(!output.is_pegout());
+
+        // Output with pushes that are e.g. OP_1 are nulldata but not pegouts
```
jonasnick: I don't see an OP_1 in the hex.

src/transaction.rs
```diff
@@ -353,7 +353,7 @@ impl TxOut {
         if iter.next() == Some(Instruction::Op(opcodes::All::OP_RETURN)) {
             for push in iter {
                 match push {
-                    Instruction::Op(op) if op as u8 > opcodes::All::OP_PUSHNUM_16 as u8 => {}
+                    Instruction::Op(op) if op as u8 > opcodes::All::OP_PUSHNUM_16 as u8 => return false,
```
jonasnick: Can we add an `Instruction::Op(op) if op as u8 <= opcodes::All::OP_PUSHNUM_16 as u8 => return true` case. I find it hard to see what isan Op or PushBytes. I think OP_1 aka OP_PUSHNUM_1 is not PushBytes and therefore would result in false here.

apoelstra: Opcodes 0 through 76 are "push this many bytes", followed by the 3 PUSHDATA opcodes, a couple reserved ones, then the numbers. Everything above OP_16 is control flow and OP_RETURN. OP_PUSHNUM_1 0x51 is definitely belowe OP_PUSHNUM_16 0x60.

I copied this logic from https://github.com/bitcoin/bitcoin/blob/master/src/script/script.cpp#L243 btw.

src/transaction.rs
apoelstra: Oh, oops, it's actually pushing a literal 0x01. Good call, this is not what I meant to test.

src/transaction.rs
apoelstra: Oh, I understand what you're saying! Will fix.

apoelstra:
Fixed @jonasnick comments. The logic should actually be correct now :)

jonasnick:
ACK

PR #17, opened 2019-04-25
redefine "pegout" to better match elementsd RPC

apoelstra:


stevenroose:
LGTM. I built hal-elements with this.
`hal elements tx decode` for the second test vector tx panic'ed with `Execution failed: internal error: entered unreachable code` while with this PR, it shows a pegout-free transaction response.

apoelstra:
Rebased; changed version to 0.7.1 rather than 0.6.1

roconnor-blockstream:
LGTM

instagibbs:
Appears to sync with elementsd. utACK

roconnor-blockstream:
See also https://github.com/ElementsProject/elements/issues/598#issuecomment-487125510

apoelstra:
Rebased.

apoelstra:
Looks good I believe with https://github.com/ElementsProject/elements/pull/610 both rust-elements and elements should agree.

src/transaction.rs
```diff
@@ -400,23 +404,37 @@ impl TxOut {
 
         // Parse destination scriptpubkey
         let script_pubkey = if let Some(Instruction::PushBytes(data)) = iter.next() {
-            Script::from(data.to_owned())
+            if data.is_empty() {
```
roconnor-blockstream: Elements was modified to make 0 byte pushes legal @ https://github.com/ElementsProject/elements/pull/610/files#diff-f7ca24fb80ddba0f291cb66344ca6fcbR222

So this block needs to be reverted now.

instagibbs: not merged yet! but looks like ACKs from relevant parties

roconnor-blockstream: oh indeed.

roconnor-blockstream: 610 is now merged and this needs to be reverted now.

apoelstra: https://github.com/ElementsProject/rust-elements/issues/20

PR #19, opened 2019-04-28
Add fast_merkle_root and AssetId type

stevenroose:
Based on Elements' [ComputeFastMerkleRoot](https://github.com/ElementsProject/elements/blob/master/src/primitives/txwitness.cpp#L50).


shesek:
Integrated into the registry project I'm working on, the implementation does seem correct now.

PR #22, opened 2019-07-24
Use secp256k1::PublicKey instead of bitcoin::PublicKey

stevenroose:


apoelstra:
concept nack. secp public keys do not round trip. You can parse a 66-byte key and it will serialize as 33 bytes.

PR #23, opened 2019-08-15
add block parsing support for dynafed

apoelstra:


src/block.rs
```diff
@@ -352,8 +458,35 @@ mod tests {
             "bcc6eb2ab6c97b9b4590825b9136f100b22e090c0469818572b8b93926a79f28"
         );
         assert_eq!(block.header.version, 0x20000000);
-        assert_eq!(block.header.proof.challenge.len(), 1 + 3 * 34 + 2);
-        assert_eq!(block.header.proof.solution.len(), 144);
+        if let ExtData::BlockProof(ref proof) = block.header.ext {
+            assert_eq!(proof.challenge.len(), 1 + 3 * 34 + 2);
+            assert_eq!(proof.solution.len(), 144);
+        } else {
+            panic!("dynafed block should not be dynafed");
```
RCasatta: "Liquid v1-style static `signblockscript block" should not be dynafed ?

src/block.rs
```diff
@@ -47,23 +93,81 @@ pub struct BlockHeader {
     /// Block height
     pub height: u32,
     /// Block height
```
stevenroose: You seem to have copied this comment from above..

src/block.rs
```diff
@@ -47,23 +93,81 @@ pub struct BlockHeader {
...
     fn bitcoin_hash(&self) -> sha256d::Hash {
         use bitcoin::consensus::Encodable;
 
+        let version = if let ExtData::Dynafed { .. } = self.ext {
+            self.version | 0x8000_0000
+        } else {
+            self.version
+        };
+
         // Everything except proof.solution goes into the hash
```
stevenroose: Maybe update this to "everything except the signblock witness goes into the hash".

src/block.rs
apoelstra: How about "dynafed test vector was parsed as non-dynafed"?

src/block.rs
apoelstra: Lol, I noticed that and thought to fix it ... but apparently did not.

apoelstra:
Addressed all nits; and added a serde-pinning commit

apoelstra:
The serde-pinning commit is not correct (it should pin `serde-derive` rather than `serde`). But probably we should just not have it; better to just keep the macros.

apoelstra:
Added manual `serde` trait implementations.

These don't match the Core JSON serialization of block headers; I'm not sure that this is a reasonable goal, given that the Core serialization includes a bunch of data (e.g. next block header) that we don't have access to, and it also does stuff like asm-decoding scripts, which is difficult for us.

PR #24, opened 2019-08-16
Add SLIP-0077

stevenroose:
An implementation of https://github.com/satoshilabs/slips/blob/master/slip-0077.md that is compatible with Elements Core and libwally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,8 @@ json-contract = [ "serde_json" ]
 
 [dependencies]
 bitcoin = "0.23"
+slip21 = "0.2.0"
+
 # While this dependency is included in bitcoin, we need this to use the macros.
 # We should probably try keep this one in sync with the bitcoin version,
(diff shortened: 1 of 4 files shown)


RCasatta:
gdk electrum will need this, at the moment is done internally with wally bindings.

Do you want to add some more test vectors?
https://github.com/ElementsProject/libwally-core/blob/721b67c061b38bc5ec2c8e5cd41e89357701685d/src/test/test_confidential_addr.py#L12
(ported to rust version https://github.com/Blockstream/gdk/blob/gdk-electrum/subprojects/gdk_rust/gdk_common/src/wally/mod.rs#L504)

stevenroose:
I added the test. I also added `Address::to_confidential` and `Address::to_unconfidential`. I hope they can just be added in this PR without need for a separate one.

RCasatta:
ack f0bc17c

PR #25, opened 2019-08-20
Update to rust-bitcoin v0.19.1, add encode module

stevenroose:
This adds an `Encodable` and `Decodable` trait for Elements types. It implements those traits for some rust-bitcoin types using the `bitcoin::consensus::Encodable` counterpart of those types.

This was needed because rust-bitcoin no longer implements `Encodable` generically for `Vec<T>`, so we couldn't encode vectors of our own types without making newtypes for them. And it's probably better like this anyway. Fairly little code duplication.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,19 +10,13 @@ documentation = "https://docs.rs/elements/"
 
 [features]
 "serde-feature" = [
-    "bitcoin/serde",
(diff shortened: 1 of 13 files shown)


stevenroose:
Closing this in favor of https://github.com/ElementsProject/rust-elements/pull/26 because of deployment based on it.

PR #26, opened 2019-08-20
Update to rust-bitcoin v0.19.1, add encode module

stevenroose:


src/encode.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    };
+}
+impl_upstream!(u8);
+impl_upstream!(u32);
+impl_upstream!(u64);
+impl_upstream!([u8; 32]);
+impl_upstream!(Vec<u8>);
+impl_upstream!(Vec<Vec<u8>>);
+impl_upstream!(btcenc::VarInt);
```
apoelstra: I think we should `pub use bitcoin::VarInt` in `lib.rs`with_capacity

src/encode.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        impl Decodable for Vec<$type> {
+            #[inline]
+            fn consensus_decode<D: io::Read>(mut d: D) -> Result<Self, Error> {
+                let len = btcenc::VarInt::consensus_decode(&mut d)?.0;
+                let byte_size = (len as usize)
+                    .checked_mul(mem::size_of::<$type>())
+                    .ok_or(self::Error::ParseFailed("Invalid length"))?;
+                if byte_size > btcenc::MAX_VEC_SIZE {
```
apoelstra: and we should `pub use bitcoin::consensus::encode::MAX_VEC_SIZE` in this module

src/encode.rs
stevenroose: Why? How about `Script`?

src/encode.rs
stevenroose: Ah yeah instead of redefine. Agree with that.

src/encode.rs
apoelstra: An elements script is not the same as bitcoin script :). But an elements varint is a bitcoin varint.

apoelstra: At some point we should have a separate `elements::script` and `elements::opcodes` modules that are independent of the Bitcoin ones, but for now (especially with our heavy Miniscript usage which is Bitcoin-only) it's fine to keep using `bitcoin::Script` in elements data structures, even though this isn't very well typed.

src/encode.rs
```diff
@@ -22,6 +22,9 @@ use ::bitcoin::consensus::encode as btcenc;
 
 use transaction::{Transaction, TxIn, TxOut};
 
+pub use ::bitcoin::consensus::encode::MAX_VEC_SIZE;
```
apoelstra: Can we `pub use` this at the top level so that users can import `elements::VarInt`?

apoelstra:
fuzztests are not finishing. I kicked the CI job. Hopefully this does not represent a perf loss. (It shouldn't; the new encoding code should have much *better* performance.)

jonasnick:
Hmm travis still not happy

src/encode.rs
stevenroose: Fixed this.

stevenroose:
I'm confused with Travis, can't really see what changed except the upstream encoding logic. But like Andrew says, it should be more efficient not less. I had to force push, let's see if it changes anything.

apoelstra:
I can reproduce the massive slowdown (like 20x) in the `deserialize_block` fuzzer locally. But it's hard to track something like this down.

apoelstra:
I think the issue is that we now have a bunch more variable-sized objects in the block headers.

I recommend just dropping the number of fuzz iterations in travis by a factor 10.

PR #31, opened 2019-09-20
Add BlockHeader::clear_witness and ::is_dynafed methods

stevenroose:


stevenroose:
Fixed for 1.22.

stevenroose:
Those matches over complex enum variants are super painful and inflated line-wise. I think we would have better have had 2 structs and simpler tuple variants taking that one struct instead of the variant taking fields. 
It's basically a PITA when you're actually using these variants to get any fields from them once you know which variant is chosen. Same goes for https://github.com/ElementsProject/rust-elements/pull/34.

So just some simple utilities to save lines and improve readability of code using these enums.

jonasnick:
Sorry, should have been more specific. I get that it's more concise but I was wondering where `clear_witness` is useful.

instagibbs:
@jonasnick witness-stripped blockheaders for when you don't need witness data to persist after witness is validated

apoelstra:
Needs rebase.

stevenroose:
Those matches over complex enum variants are super painful and inflated line-wise. I think we would have better have had 2 structs and simpler tuple variants taking that one struct instead of the variant taking fields. 
It's basically a PITA when you're actually using these variants to get any fields from them once you know which variant is chosen. Same goes for https://github.com/ElementsProject/rust-elements/pull/34.

So just some simple utilities to save lines and improve readability of code using these enums.

PR #32, opened 2019-09-23
Fix encoding bug (length return value of confidential objects)

apoelstra:


stevenroose:
Good catch!

apoelstra:
Rebased. Probably will backport to 0.8.0 though.

PR #34, opened 2019-09-25
Add some inspector methods to dynafed::Params + support root calculation

stevenroose:
The Rust enum stuff is all fancy and such, but it's often a real pain in the ass to either just check which of the vairants is used or to get a field out if you already know what the variant is.

### DIFF
--- a/src/dynafed.rs
+++ b/src/dynafed.rs
@@ -17,6 +17,7 @@
 use std::io;
 
 use bitcoin;
+use bitcoin::hashes::{Hash, sha256, sha256d};
 #[cfg(feature = "serde")] use serde::{Deserialize, Deserializer, Serialize, Serializer};
 #[cfg(feature = "serde")] use std::fmt;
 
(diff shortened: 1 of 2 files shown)


src/dynafed.rs
```diff
@@ -272,3 +369,81 @@ impl Decodable for Params {
...
+        let fp_script = vec![4];
+        let ext = vec![vec![5, 6], vec![7]];
+
+        let compact_entry = Params::Compact {
+            signblockscript: signblockscript.clone(),
+            signblock_witness_limit: signblock_wl,
+        };
+        assert_eq!(
+            compact_entry.calculate_root().to_hex(),
+            "dff5f3793abc06a6d75e80fe3cfd47406f732fa4ec9305960ae2a229222a1ad5"
```
jonasnick: This doesn't appear to match elements test hash `4587cf1c2448740c3fe4bf0d3ab22f3f193b18f43c4ee47da3690c53becb7fa4`?

stevenroose: Ah yeah I forgot to update those after I changed the test. In this PR you can see the correct ones: https://github.com/ElementsProject/elements/pull/719/files
I'll fix this in a force-push.

src/dynafed.rs
```diff
@@ -272,3 +369,81 @@ impl Decodable for Params {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use bitcoin;
+    use bitcoin::hashes::hex::ToHex;
+
+    #[test]
+    fn test_param_roots() {
+        // Taken from the following Elements Core test:
```
jonasnick: Would have been better I think to just include a github link to commit and line number. But not all that important.

stevenroose: Yeah the test in Elements was still up for review and might not be merged or be changed in the future. I figured it's not that big so better just make sure to have it.

PR #35, opened 2019-11-07
dynafed: replace `calculate_root` with new version

apoelstra:
Needed for https://github.com/ElementsProject/elements/pull/773


src/address.rs
```diff
@@ -75,7 +75,7 @@ impl fmt::Display for AddressError {
 }
 
 impl error::Error for AddressError {
-    fn cause(&self) -> Option<&error::Error> {
+    fn cause(&self) -> Option<&dyn error::Error> {
```
stevenroose: This won't work until we bump MSRV I think.

apoelstra: oh oops, I didn't mean that to be in this PR

PR #36, opened 2019-11-22
Add Params::elided_root and Params::into_compact

stevenroose:
Sorry, I should have noticed this was missing while reviewing the PR that added the elided_root field. There are getters for all the other fields, so we should probably have this one. I already bumped into missing this one.


apoelstra:
lgtm. Want to add a commit that bumps the minor version?

PR #37, opened 2020-01-09
Update to rust-bitcoin v0.23

stevenroose:


apoelstra:
Looks good to me ... but arguably this crate should have its own blockhash/txid types to differentiate elements objects from bitcoin ones.

apoelstra:
Gonna go ahead and merge this. We need to have a wider discussion about how bitcoin-specific the types representing pegin data should be (in principle we can support arbitrary other blockchains) and also how separate the Bitcoin and Elements types should be.

PR #43, opened 2020-03-24
Add Transaction::get_size()

shesek:
Useful to compute the size of a transaction without having to re-serialize it.

Also see: https://github.com/rust-bitcoin/rust-bitcoin/pull/416.


stevenroose:
Travis and GitHub seem to be having issues so GH doesn't recognize the build passed. I can't merge manually either. I tried restarting a single job, hope that works.

PR #44, opened 2020-03-25
Add Block::get_size() and Block::get_weight()

shesek:
Similarly to https://github.com/rust-bitcoin/rust-bitcoin/pull/417.

Unlike https://github.com/rust-bitcoin/rust-bitcoin/pull/417, this doesn't avoid serialization entirely and still serializes the header in order to determine its size. This could be avoided too, but most of the performance gain is already achieved by not serializing transactions, which seemed good enough.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -19,14 +19,14 @@ use std::io;
 
 use bitcoin;
 use bitcoin::blockdata::script::Script;
-use bitcoin::{BitcoinHash, BlockHash};
+use bitcoin::{BitcoinHash, BlockHash, VarInt};
 use bitcoin::hashes::{Hash, sha256};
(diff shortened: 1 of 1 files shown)


shesek:
I also verified the weight and size to match what cpp-elements reports for the [liquid mainnet block at height 2](https://blockstream.info/liquid/block/57f26b5a65e461e8aa690e97595702605ef4fff5308d9507121b44e035663dcd).

stevenroose:
This needs a rebase.

shesek:
@stevenroose rebased.

stevenroose:
closing and reopening to try to get Travis to do something

stevenroose:
@jonasnick @apoelstra can any of you merge this? I checked Travis manually and all jobs passed, is just that the reporting back to Github seems broken

PR #45, opened 2020-04-03
Replace Transaction::fee with ::fee_in and ::all_fees 

stevenroose:
@RCasatta pointed out that we don't check the asset type on fee outputs. In theory they can be any asset.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -16,14 +16,16 @@
 //!
 
 use std::{io, fmt};
+use std::collections::HashMap;
 
 use bitcoin::{self, BitcoinHash, Txid, VarInt};
 use bitcoin::blockdata::opcodes;
 use bitcoin::blockdata::script::{Script, Instruction};
-use bitcoin::hashes::Hash;
+use bitcoin::hashes::{Hash, sha256, sha256d};
 
 use confidential;
 use encode::{self, Encodable, Decodable};
+use issuance::AssetId;
 
 /// Description of an asset issuance in a transaction input
 #[derive(Copy, Clone, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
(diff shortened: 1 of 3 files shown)


stevenroose:
As a follow-up I'd like to change the internal type for `confidential::Asset` from `sha256d::Hash` which is just wrong to `AssetId`.

src/transaction.rs
```diff
@@ -613,10 +615,25 @@ impl Transaction {
         bitcoin::Wtxid::from_engine(enc)
     }
 
-    /// Get the total transaction fee.
-    pub fn fee(&self) -> u64 {
+    /// Get the total transaction fee in the given asset.
+    pub fn fee_in(&self, asset: AssetId) -> u64 {
```
RCasatta, line 619: maybe simpler calling all_fees() ?

src/transaction.rs
```diff
@@ -613,10 +615,25 @@ impl Transaction {
...
+            .filter(|o| o.is_fee() && o.asset == asset)
+            .filter_map(|o| o.value.explicit())
+            .sum()
+    }
+
+    /// Get all fees in all assets.
+    pub fn all_fees(&self) -> HashMap<AssetId, u64> {
+        let mut fees = HashMap::new();
+        for out in self.output.iter().filter(|o| o.is_fee()) {
+            let asset = AssetId::from_inner(sha256::Midstate::from_inner(out.asset.explicit().unwrap().into_inner()));
```
RCasatta, line 632: maybe a comment why is safe to unwrap() ?

src/transaction.rs
stevenroose: I'd like to avoid building the hashmap etc.

src/transaction.rs
stevenroose: good point, I have that comment in fee_in, but not here, will add.

RCasatta:
ack 16025626253f5c3414ab4d584fa6d0e041b12b7d

stevenroose:
Travis is acting out, this passed CI here: https://travis-ci.org/github/ElementsProject/rust-elements/builds/670508921.

Issue #18, opened 2019-04-25
confidential value prefix bytes should be enforced when decoding


apoelstra:
The `confidential::` types all allow any prefix. We shouldn't allow that.

roconnor-blockstream:
Is this fixed by #47?

apoelstra:
Yes, good catch

PR #47, opened 2020-04-04
Confidential improvements

stevenroose:
Solves https://github.com/ElementsProject/rust-elements/issues/18: enforce correct commitment prefixes.

Also replaces the `confidential::Nonce` inner value with `[u8; 32]` because it's not a `sha256d::Hash`. I'm not sure if this will cause any problem anywhere because they are not displayed the same way.

Plus some other improvements.

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -104,22 +104,6 @@ macro_rules! impl_confidential_commitment {
             }
         }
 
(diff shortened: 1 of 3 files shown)


apoelstra:
Will require a major version bump since it changes the type of the `Confidential` variant of the public `Nonce` enum

PR #48, opened 2020-04-05
Add ContractHash type to ensure reverse hex serialization

stevenroose:
Which is what Elements Core expects.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,10 @@ documentation = "https://docs.rs/elements/"
 
 [dependencies]
 bitcoin = "0.23"
+# While this dependency is included in bitcoin, we need this to use the macros.
+# We should probably try keep this one in sync with the bitcoin version,
+# to avoid requiring two version of bitcoin_hashes.
+bitcoin_hashes = "0.7.6"
 
 [dependencies.serde]
 version = "1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -17,6 +17,10 @@ documentation = "https://docs.rs/elements/"
 
 [dependencies]
 bitcoin = "0.23"
+# While this dependency is included in bitcoin, we need this to use the macros.
+# We should probably try keep this one in sync with the bitcoin version,
+# to avoid requiring two version of bitcoin_hashes.
+bitcoin_hashes = "0.7.6"
```
apoelstra, line 23: Looks like rust-bitcoin is using >0.7.3 ... I think this is okay. Technically it means that somebody could pin their version at 0.7.3 and cargo wouldn't complain, but this seems like a pretty unlikely scenario.

apoelstra: Ah @stevenroose says elsewhere that we need this to access the macros. Can you update the comment to reflect this?

stevenroose: The first line of the comment says we need this to access the macros :)

If someone pinned their version to 0.7.3, he wouldn't be able to build elements.

stevenroose:
Fixed the unit test and rebased.

shesek:
Nice, that's pretty useful, thanks! I used this for esplora/electrs and for the liquid asset registry, works for me as well.

PR #49, opened 2020-04-07
Add AssetId::from_slice() convenience method

shesek:
To match the one found on other hash types.


stevenroose:
You have a compilation failure on Rust 1.22.

stevenroose:
Otherwise concept ACK.

shesek:
@stevenroose Good catch! Fixed.

PR #50, opened 2020-04-21
Fix and export ContractHash

stevenroose:


stevenroose:
Heh, I force pushed another commit but GitHub is not showing it.

shesek:
Wasn't ContractHash using sha256d to get reverse serialization that matches cpp-elements?

stevenroose:
Yeah it it's not a double hash. So then I fixed that with the macro that allows creating a hash newtype but changing the reversed-ness :) It's messy, but well. It's more important that `ContractHash::hash` does the right thing.

shesek:
Ah, nice! This is of course better. :) I guess this what the last `true` macro argument signifies?

stevenroose:
Yeah. It's a but bad that it's just a boolean, right? Perhaps I should make that a `reverse: $rev:expr` instead in the macro definition...

PR #51, opened 2020-04-21
Add support for hashing JSON asset contract

stevenroose:
Annoyingly this takes 2 extra dependencies. I made them optional, but default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,9 @@ repository = "https://github.com/ElementsProject/rust-elements/"
 documentation = "https://docs.rs/elements/"
 
 [features]
+default = [ "json-contract" ]
+
+json-contract = [ "serde_json" ]
 "serde-feature" = [
     "bitcoin/use-serde",
     "serde"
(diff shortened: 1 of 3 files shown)


RCasatta:
You can use std collection `BTreeMap` to avoid using IndexMap dep

stevenroose:
@RCasatta hmm, because the BTreeMap is automatically ordered. Hmm. Yeah I used indexmap because that's what serde_json uses internally for maps which "preserve order". But perhaps if the order needs to be lexicographic anyway, a BTreeMap would make sense. I don't know the algorithm, but it would do some sorting/comparing on every insert, though. But I suppose to be able to drop a dependency, that would make sense.

stevenroose:
Hmm, I'm confused now. I was quite sure that both our methods would break if I introduced new fields in a nested object. Like `{"entity":{"domain":"xxx","author":"yyy"}}`. But somehow it doesn't; I'm not sure if I'm just lucky.

stevenroose:
Thanks for the suggestion, btw! It seems that serializing into a BTreeMap somehow magically also orders nested objects...

stevenroose:
Cool, you can see here that serde's default implementation for `BTreeMap` works recursively. I'm impressed :)

RCasatta:
ack 94ba7f6

PR #53, opened 2020-05-21
Some minor cleanups in Transaction code

stevenroose:
This is a breaking change so it better be deferred until we're doing a breaking change.

I'm starting to think that it'd be a good idea to have separate elements-specific hash newtypes to distinguish `bitcoin::BlockHash` and `elements::BlockHash` etc. I might propose this in another PR.

### DIFF
--- a/src/issuance.rs
+++ b/src/issuance.rs
@@ -17,8 +17,7 @@
 use std::io;
 use std::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


apoelstra:
Closing, I think this is all superceded by #57 and #59

PR #59, opened 2020-09-29
add elements-specific `Script` and `opcodes` types

apoelstra:
needed for CHECKSIGFROMSTACK

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -154,6 +154,7 @@ impl_upstream!(u8);
 impl_upstream!(u32);
 impl_upstream!(u64);
 impl_upstream!([u8; 32]);
+impl_upstream!(Box<[u8]>);
 impl_upstream!(Vec<u8>);
 impl_upstream!(Vec<Vec<u8>>);
 impl_upstream!(btcenc::VarInt);
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -158,11 +158,7 @@ impl_upstream!(Box<[u8]>);
 impl_upstream!(Vec<u8>);
 impl_upstream!(Vec<Vec<u8>>);
 impl_upstream!(btcenc::VarInt);
-impl_upstream!(::bitcoin::blockdata::script::Script);
-impl_upstream!(::bitcoin::hashes::sha256d::Hash);
-impl_upstream!(::bitcoin::Txid);
-impl_upstream!(::bitcoin::TxMerkleNode);
-impl_upstream!(::bitcoin::BlockHash);
+impl_upstream!(::hashes::sha256d::Hash);
 
 // Vectors
 macro_rules! impl_vec {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,8 @@ pub mod encode;
 mod fast_merkle_root;
 pub mod hash_types;
 pub mod issuance;
+pub mod opcodes;
+pub mod script;
 mod transaction;
 pub mod slip77;
 
@@ -56,4 +58,5 @@ pub use ::bitcoin::consensus::encode::VarInt;
 pub use fast_merkle_root::fast_merkle_root;
 pub use hash_types::*;
 pub use issuance::{AssetId, ContractHash};
+pub use script::Script;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,7 @@ pub mod confidential;
 pub mod dynafed;
 pub mod encode;
 mod fast_merkle_root;
+pub mod hash_types;
 pub mod issuance;
 mod transaction;
 pub mod slip77;
@@ -53,5 +54,6 @@ pub use block::{BlockHeader, Block};
 pub use block::ExtData as BlockExtData;
 pub use ::bitcoin::consensus::encode::VarInt;
 pub use fast_merkle_root::fast_merkle_root;
+pub use hash_types::*;
 pub use issuance::{AssetId, ContractHash};
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,7 +27,7 @@
 
 pub extern crate bitcoin;
 #[macro_use]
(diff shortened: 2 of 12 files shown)


RCasatta:
I imagine it's controversial to add an "esoteric" compile feature in rust-bitcoin with a CHECKSIGFROMSTACK operator? Or maybe not enough to avoid duplicating so much code?

Disclosure: Script struct is basically the only one I can share between bitcoin and elements in gdk-rust code

apoelstra:
@RCasatta arguably you shouldn't be sharing `Script` types between bitcoin and elements ... these are different kinds of objects. Is your concern about code size? Maybe we can shrink `elements::Script` by implementing all the methods in terms of `bitcoin::Script` ... though the unfortunately the bulk of the code is the `Instructions` iterator which has to be Elements-specific as it uses `elements::opcodes`.

But yes, there is no way we can get a CHECKSIGFROMSTACK operator into rust-bitcoin, even with a feature flag.

RCasatta:
yes, I agree, I will create an enum to discriminate between them like I am doing for other structs

src/transaction.rs
```diff
@@ -18,14 +18,15 @@
 use std::{io, fmt};
 use std::collections::HashMap;
 
-use bitcoin::{self, Txid, VarInt};
+use bitcoin::{self, VarInt};
 use bitcoin::blockdata::opcodes;
 use bitcoin::blockdata::script::{Script, Instruction};
```
sanket1729, line 23: Should be changed to script::Script instead of bitcoin::Script?

sanket1729:
Can we also add OP_RETURN_195 as OP_SUBSTRLAZY. 
https://github.com/ElementsProject/elements/blob/88067ddbe4223e43a4c86912928b28b834bbf69e/src/script/interpreter.cpp#L818 ?

apoelstra:
lol I have avoided learning what that opcode did for years

sanket1729:
Finally, can we re-export as `bitcoin`, `secp256k1`, and `hashes` library for ease of convenience?

apoelstra:
Rebased and added two commits to address your issues, and also one which fixes the opcode classification methods.

sanket1729:
Btw, this fails on serde feature which is not tested in CI.

src/encode.rs
```diff
@@ -154,6 +154,7 @@ impl_upstream!(u8);
 impl_upstream!(u32);
 impl_upstream!(u64);
 impl_upstream!([u8; 32]);
+impl_upstream!(Box<[u8]>);
```
sanket1729, line 157: I don't know if all of the `impl_upstream!` are required for `bitcoin::` types anymore.

apoelstra: Heh, good find, I removed all the ones that the compiler would let me.

sanket1729:
tAck 1446f9109efab64c6593b2a452ebe6d02897e574

RCasatta:
Concept ACK

src/lib.rs
```diff
@@ -27,7 +27,7 @@
 
 pub extern crate bitcoin;
 #[macro_use]
-pub extern crate bitcoin_hashes;
+pub extern crate bitcoin_hashes as do_not_use_me_just_imported_for_the_macros;
```
stevenroose, line 30: Lol yeah I don't think it needs to be a `pub use`, does it? Then it doesn't matter what name it has.

src/lib.rs
```diff
@@ -49,6 +49,8 @@ pub mod script;
 mod transaction;
 pub mod slip77;
 
+// re-export bitcoin deps which we re-use
+pub use bitcoin::{bech32, hashes, secp256k1};
```
stevenroose, line 53: Hmm, this is not super fun. I mean this will show up as `elements::hashes` and `elements::bech32`. Is that intended?

Alternatively, we can do 
```rust
pub mod bitcoin {
    pub use bitcoin::{bech32, hashes, secp256k1};
}
```
Hmm, but that might mess with our internal use of the `bitcoin` crate import

apoelstra: I think, just doing `pub extern crate bitcoin` would get us the effect you want.

But my intention had been that we'd have `elements::hashes`, `elements::bech32`, etc. Do you think that's too confusing?

src/lib.rs
apoelstra: loll yeah good call

src/lib.rs
sanket1729: I think conceptually, we should be using everything from elements. Only minimal code that deals with PegIn/PegOut should have access to the `bitcoin` crate. 

To this end, I would even propose re-exporting the  `bitcoin::PublicKey` to `elements::PublicKey` and other possible dependencies. Maybe this will end causing more confusion, what do you think?

apoelstra: Well, one unfortunate thing is that if we merely re-export, rustc will still treat them as the same type. So if we export `elements::PublicKey` then users will still be able to call functions that expect a `bitcoin::PublicKey`. And actualy in rustc error messages, the path that gets shown is pretty-much random as near as I can tell, which is even more confusing.

stevenroose: Yeah that's quite the same as doing `pub extern crate bitcoin;`.

I can ACK this change, though. Seems fine.

Just a gentle reminder of https://github.com/stevenroose/rust-elements (branch `elements2`) is a fork of rust-bitcoin so it actually has all equivalent `elements::Xxx` types that are distinct.

apoelstra: If we were to make rust-elements a fork of rust-bitcoin, these crates would be re-exported in the same way, and would still not be distinct from the ones exported from rust-bitcoin.

apoelstra:
@stevenroose removed the `pub` that you nitted

stevenroose (approved):
LGTM

apoelstra:
rebased due to silent merge conflict with #62

PR #64, opened 2020-11-20
Add functionality to create sighash(legacy + segwit)

sanket1729:
Builds on top of #59. Will rebase after that is merged. Let me if you any ideas about how to quickly generate test vectors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,6 +49,8 @@ pub mod script;
 mod transaction;
 pub mod slip77;
 pub mod sighash;
+// consider making upstream public
+mod endian;
 // re-export bitcoin deps which we re-use
 pub use bitcoin::{bech32, hashes, secp256k1};
 // export everything at the top level so it can be used as `elements::Transaction` etc.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,7 +53,7 @@ pub mod slip77;
 pub use bitcoin::{bech32, hashes, secp256k1};
 // export everything at the top level so it can be used as `elements::Transaction` etc.
 pub use address::{Address, AddressParams, AddressError};
-pub use transaction::{OutPoint, PeginData, PegoutData, TxIn, TxOut, TxInWitness, TxOutWitness, Transaction, AssetIssuance};
+pub use transaction::{OutPoint, PeginData, PegoutData, SigHashType, TxIn, TxOut, TxInWitness, TxOutWitness, Transaction, AssetIssuance};
 pub use block::{BlockHeader, Block};
 pub use block::ExtData as BlockExtData;
 pub use ::bitcoin::consensus::encode::VarInt;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@ pub mod opcodes;
 pub mod script;
 mod transaction;
 pub mod slip77;
-
+pub mod sighash;
 // re-export bitcoin deps which we re-use
 pub use bitcoin::{bech32, hashes, secp256k1};
 // export everything at the top level so it can be used as `elements::Transaction` etc.
--- a/src/sighash.rs
+++ b/src/sighash.rs
@@ -1,0 +1,253 @@
+// Rust Bitcoin Library
+// Written in 2018 by
+//     Andrew Poelstra <apoelstra@wpsoftware.net>
+// To the extent possible under law, the author(s) have dedicated all
+// copyright and related and neighboring rights to this software to
+// the public domain worldwide. This software is distributed without
+// any warranty.
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+//! BIP143 Implementation
+//!
+//! Implementation of BIP143 Segwit-style signatures. Should be sufficient
+//! to create signatures for Segwit transactions (which should be pushed into
+//! the appropriate place in the `Transaction::witness` array) or bcash
+//! signatures, which are placed in the scriptSig.
+//!
+
+use encode::{self, Encodable};
+use hash_types::SigHash;
+use hashes::{sha256d, Hash};
+use script::Script;
+use std::ops::Deref;
+use std::io;
+use transaction::SigHashType;
+use transaction::Transaction;
+use confidential;
+
+/// A replacement for SigHashComponents which supports all sighash modes
+pub struct SigHashCache<T> {
+    /// Access to transaction required for various introspection
+    tx: T,
+    /// Hash of all the previous outputs, computed as required
+    hash_prevouts: Option<sha256d::Hash>,
+    /// Hash of all the input sequence nos, computed as required
+    hash_sequence: Option<sha256d::Hash>,
+    /// Hash of all the outputs in this transaction, computed as required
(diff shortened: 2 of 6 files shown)


apoelstra:
can rebase now

apoelstra:
We probably shouldn't call this "bip143" since it does something a bit different and weirder than BIP143.

Also worth considering how we can fit legacy and taproot sighashes into the same `SigHashCache` model. Or even keep the same type.

sanket1729:
> We probably shouldn't call this "bip143" since it does something a bit different and weirder than BIP143. ?

We can call it segwit_elements, or segwit?

> Also worth considering how we can fit legacy and taproot sighashes into the same SigHashCache model. Or even keep the same type.

I think all the signature hashes(taproot and further upgrades) should conceptually fit into the `SigHashCache`. It is after all a cache of all transaction information required for a signature. After which we can have methods on it with different names, say `sighash_legacy`, `sighash_segwitv0`, `sighash_taproot` etc...

sanket1729:
I have also added another commit to add some new functions that were missing in #59(Added to rust-bitcoin afterwards) but would make it more convenient to implement elements-miniscript downstream.

sanket1729:
I starting to think that there should be only one signature hash cache for all the types of sighash(legacy, witness, taproot) because when verifying multiple inputs of the transaction containing different sighashes scheme, it would useful to reuse the cache data instead of computing it for a separate type.

apoelstra:
Great. As for the Taproot hash, I will PR a .md document to upstream elements/ that we can implement against. And maybe actually implement it in Elements :).

apoelstra:
I wonder if there is anything we can cache for the legacy sighash? Maybe a sha2 midstate/engine?

sanket1729:
It is not easy to cache the state for legacy because it is dependant on the index of the transaction. The only thing we can cache is midstate up to the point of first index script sig. So, I think it would be better to do in the way rust-bitcoin does it. Have a method for `legacy_signature_hash` in `transcation.rs` and other sighash methods in util.

thomaseizinger:
libwally uses the value commitment for computing the transaction hash of confidential transactions: https://github.com/ElementsProject/libwally-core/blob/319fd1461a85dcfdd7566d4cd78e36fb764b03f9/include/wally_transaction.h#L1148

Will this `SigHashCache` (eventually) also work for confidential transactions?

sanket1729:
Yes, the currently implemented `SigHashCache`(segwit style sighash) in this PR will work for confidential transactions too. I also plan to add a commit to this PR to support signature_hash for legacy CT.

sanket1729:
Renamed to the module to `segwit_sighashes`. Let me know if you can think of a better name. We can use this same `SigHashCache` in this module for taproot too. Cross checked test vectors covering all sighashtypes and issuances from elements python test framework.

sanket1729:
This PR now has support for creating Legacy and Segwit sighashes with confidential values. Cross tested across elements.

apoelstra:
> It is not easy to cache the state for legacy because it is dependant on the index of the transaction. The only thing we can cache is midstate up to the point of first index script sig. So, I think it would be better to do in the way rust-bitcoin does it. Have a method for `legacy_signature_hash` in `transcation.rs` and other sighash methods in util.

Let's expose this method on `SigHashCache` though as well for the sake of having a consistent API, even if nothing is cached.

src/segwit_sighashes.rs
```diff
@@ -0,0 +1,236 @@
...
+use hashes::{sha256d, Hash};
+use script::Script;
+use std::ops::Deref;
+use std::io;
+use transaction::SigHashType;
+use transaction::Transaction;
+use confidential;
+
+/// A replacement for SigHashComponents which supports all sighash modes
+pub struct SigHashCache<R: Deref<Target = Transaction>> {
```
thomaseizinger, line 33: Given that you are not using any associated types of the trait bound in the field declarations, you could consider removing it from the struct declaration altogether.
All the useful functionality of `SigHashCache` is guarded by the trait bound on the `impl` block anyway.

Having trait bounds on struct declarations can be painful further down the line because they have to be restated everywhere you use the struct, even if you are not using the functionality. It might not matter in this case but it is still unnecessary to include here :)

src/transaction.rs
```diff
@@ -629,6 +631,120 @@ impl Transaction {
...
+    /// # Panics Panics if `input_index` is greater than or equal to `self.input.len()`
+    ///
+    pub fn encode_signing_data_to<Write: io::Write>(
+        &self,
+        mut writer: Write,
+        input_index: usize,
+        script_pubkey: &Script,
+        sighash_type: SigHashType,
+    ) -> Result<(), encode::Error> {
+        assert!(input_index < self.input.len());  // Panic on OOB
```
thomaseizinger, line 652: This function is already fallible. Why do we need to panic here?

thomaseizinger: I am worried that, for example, in the case of a multi-party protocol, either the transaction or the index is coming from another party and hence we cannot always guarantee this pre-condition. Having this panic here would mean that we need to explicitly guard against this outside of this function to avoid other parties being able to panic our code.

sanket1729: The main function which parties should be using `signature_hash()` is not fallible and we are using `expect` over there when we call this function inside it. But you are right, there is no reason this particular function should panic.

src/segwit_sighashes.rs
sanket1729: This reasoning makes sense. I blindly adapted the `SigHashCache` code from rust-bitcoin to rust-elements. Maybe @apoelstra knows why have the bound.

apoelstra: No, `SigHashCache` was proposed and merged without my knowledge or intervention :)

I'm a little surprised that we don't use the bound anywhere ... how do we compute hashes of transaction data without knowing that we have a transaction? But we can remove it if this is true.

thomaseizinger: > I'm a little surprised that we don't use the bound anywhere ... how do we compute hashes of transaction data without knowing that we have a transaction? But we can remove it if this is true.

The bound is used on the `impl` block!
But you don't need the bound on the `struct` declaration to be able to put a bound on the `impl` block.

That is how most of the stdlib is organized: structs have "plain" type parameters and impl blocks then constrain, which functionality is available given which bounds.

The only exceptions to this are:

- Bounds you need in the `Drop` implementation, those will need to go on the struct as well
- Associated types used to declare more fields as in:

```rust
struct Foo<T: Bar> {
	baz: T::Baz
}
```

src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub mod opcodes;
 pub mod script;
 mod transaction;
 pub mod slip77;
-
```
apoelstra, line 51: would prefer you retain this whitespace

src/sighash.rs
```diff
@@ -0,0 +1,253 @@
...
+            for txout in output {
+                txout.consensus_encode(&mut enc).unwrap();
+            }
+            sha256d::Hash::from_engine(enc)
+        })
+    }
+
+    /// Encode the BIP143 signing data for any flag type into a given object implementing a
+    /// std::io::Write trait.
+    ///
```
apoelstra, line 234: trailing whitespace

apoelstra (approved):
ack 9d76ae6c196edc3ead7df746334173af899b6519

PR #70, opened 2021-02-04
Model confidential and explicit transaction outputs separately

### LINKED ISSUE #66: About contributing back a fairly big change
Hey everyone,

we've vendored this library [here](https://github.com/comit-network/droplet/tree/master/elements-fun) and made some bigger changes across the board, most notably:

- Model `TxOut`s distinctly as explicit or confidential instead of each individual field: https://github.com/comit-network/droplet/blob/8cb45d3029c4edcf499ae8cc21c98e43829a6d59/elements-fun/src/transaction.rs#L79-L90
This model makes expressing nonsensical states like a confidential value with an explicit asset impossible. We've found that quite helpful in trying to wrap our heads around elements. It is quite possible that some edge-cases are not covered but for now all the tests pass.
- Direct integration with our [fork](https://github.com/comit-network/rust-secp256k1-zkp) of `rust-secp256k1` that binds against `libsecp256k1-zkp` to provide all the cryptographic primitives for blinding / unblinding assets etc. This allows us to not have a dependency on `libwally` (which doesn't compile to WASM and has kind of quirky APIs). This is not fully complete (see this [PR](https://github.com/comit-network/droplet/pull/14)) because we need [`SigHashCache`](https://github.com/ElementsProject/rust-elements/pull/64) to land on master here :D 

For now, all of this happens within a single cargo workspace so we don't have to bother with releases. However, we are keen to upstream some of these changes. In particular:

1. Could our [fork](https://github.com/comit-network/rust-secp256k1-zkp) of `rust-secp256k1` potentially replace https://github.com/ElementsProject/rust-secp256k1-zkp? I'd assume we would have to go through some bike-shedding of the APIs as we've currently optimized ours for easy of understanding (YMMV) and to prevent misuse as much as possible at the cost of doing extra allocations for example.
2. The already mentioned refactoring of how `Transaction`s are modeled might be of interest. Similarly, it would take some effort to untangle the changes made there to make them upstreamable because we for example didn't adhere to a MSRV of 1.29.

For example, something that would be possible if we can get (1) done is:

- release a new version of `secp256k1-zkp` to crates.io (it seems that the name `secp256k1-zkp` is pointing to https://github.com/ElementsProject/rust-secp256k1-zkp)
- depend on it in `rust-elements`
- upstream changes like [`ConfidentialTxOut::unblind()`](https://github.com/comit-network/droplet/blob/54f9f4bda551a35d210ee14a57517cc060d47552/elements-fun/src/transaction.rs#L1296-L1333) which provides a quite straight-forward API to invoke some decently heavy cryptography.

Directly depending on `secp256k1-zkp` also removes a lot of duplicated knowledge like how to parse and serialize generators, pedersen commitments, proofs, etc. Additionally, it allows us to use all this cryptography from [WASM](https://github.com/comit-network/droplet/tree/master/waves) which is why we are going through all of this :)

Thoughts?

### LINKED ISSUE #67: Modelling confidential vs explicit transaction outputs
As mentioned [here](https://github.com/ElementsProject/rust-elements/issues/66#issuecomment-744080927), this is a dedicated issue for discussing our attempt at re-modelling confidential and explicit transactions.

We assumed that a `TxOut` can either confidential or explicit - i.e. both the asset and the value are either confidential or explicit - but this turned out to [not be true](https://github.com/ElementsProject/rust-elements/issues/66#issuecomment-743783437).

To be honest, I don't fully understand what a "pre-asset" confidential transaction is? Can I read up on this somewhere to get an overview of what the usecases are and what they look like on-chain?

luckysori:
These are most (all?) of the relevant changes we introduced in our fork of this library (found here: https://github.com/comit-network/waves/tree/master/elements-fun).

The motivation for this PR is to allow users of this library to construct valid confidential transactions, something that used to only be feasible (but quite cumbersome) by also depending on `lib-wally` via https://github.com/RCasatta/wally/.

Closes #66.
Closes #67.
Closes https://github.com/comit-network/waves/issues/130.

CC @thomaseizinger.

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -155,7 +155,7 @@ impl Decodable for Value {
                 let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
                 Ok(Value::Confidential(PedersenCommitment::from_slice(&bytes)?))
             }
-            p => return Err(encode::Error::InvalidConfidentialPrefix(p)),
+            p => Err(encode::Error::InvalidConfidentialPrefix(p)),
         }
     }
 }
@@ -195,7 +195,7 @@ impl<'de> Deserialize<'de> for Value {
                 f.write_str("a committed value")
             }
 
-            fn visit_seq<A: SeqAccess<'de>>(self, mut access: A) -> Result<Self::Value, A::Error> {
+            fn visit_seq<A: SeqAccess<'de>>(self, mut access: A) -> Result<Value, A::Error> {
                 let prefix: u8 = if let Some(x) = access.next_element()? {
                     x
                 } else {
@@ -218,12 +218,10 @@ impl<'de> Deserialize<'de> for Value {
                         }
                         None => Err(A::Error::custom("missing commitment")),
                     },
-                    p => {
-                        return Err(A::Error::custom(format!(
-                            "invalid commitment, invalid prefix: 0x{:02x}",
-                            p
-                        )))
-                    }
+                    p => Err(A::Error::custom(format!(
+                        "invalid commitment, invalid prefix: 0x{:02x}",
+                        p
+                    ))),
                 }
             }
         }
@@ -361,7 +359,7 @@ impl Decodable for Asset {
                 let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
                 Ok(Asset::Confidential(Generator::from_slice(&bytes)?))
             }
-            p => return Err(encode::Error::InvalidConfidentialPrefix(p)),
+            p => Err(encode::Error::InvalidConfidentialPrefix(p)),
         }
     }
 }
@@ -423,12 +421,10 @@ impl<'de> Deserialize<'de> for Asset {
                         }
                         None => Err(A::Error::custom("missing commitment")),
                     },
-                    p => {
-                        return Err(A::Error::custom(format!(
-                            "invalid commitment, invalid prefix: 0x{:02x}",
-                            p
-                        )))
-                    }
+                    p => Err(A::Error::custom(format!(
+                        "invalid commitment, invalid prefix: 0x{:02x}",
+                        p
+                    ))),
                 }
             }
         }
@@ -577,7 +573,7 @@ impl Decodable for Nonce {
                     PublicKey::from_slice(&bytes).map_err(secp256k1_zkp::Error::Upstream)?,
                 ))
             }
-            p => return Err(encode::Error::InvalidConfidentialPrefix(p)),
+            p => Err(encode::Error::InvalidConfidentialPrefix(p)),
         }
     }
 }
@@ -639,12 +635,10 @@ impl<'de> Deserialize<'de> for Nonce {
                         }
                         None => Err(A::Error::custom("missing commitment")),
                     },
-                    p => {
-                        return Err(A::Error::custom(format!(
-                            "invalid commitment, invalid prefix: 0x{:02x}",
-                            p
-                        )))
-                    }
+                    p => Err(A::Error::custom(format!(
+                        "invalid commitment, invalid prefix: 0x{:02x}",
+                        p
+                    ))),
                 }
             }
         }
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -17,7 +17,13 @@
 //! Structures representing Pedersen commitments of various types
 //!
 
-use secp256k1_zkp::{self, Generator, PedersenCommitment, PublicKey};
+use bitcoin::hashes::{sha256d, Hash};
+use secp256k1_zkp::{
+    self, compute_adaptive_blinding_factor,
+    ecdh::SharedSecret,
+    rand::{CryptoRng, Rng, RngCore},
+    CommitmentSecrets, Generator, PedersenCommitment, PublicKey, Secp256k1, SecretKey, Signing,
+};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
@@ -38,6 +44,16 @@ pub enum Value {
 }
 
 impl Value {
+    /// Create value commitment.
+    pub fn new_confidential<C: Signing>(
+        secp: &Secp256k1<C>,
+        value: u64,
+        asset: Generator,
+        bf: ValueBlindingFactor,
+    ) -> Self {
+        Value::Confidential(PedersenCommitment::new(secp, value, bf.0, asset))
+    }
+
     /// Serialized length, in bytes
     pub fn encoded_length(&self) -> usize {
         match *self {
@@ -242,6 +258,19 @@ pub enum Asset {
 }
 
 impl Asset {
+    /// Create asset commitment.
+    pub fn new_confidential<C: Signing>(
+        secp: &Secp256k1<C>,
+        asset: AssetId,
+        bf: AssetBlindingFactor,
+    ) -> Self {
+        Asset::Confidential(Generator::new_blinded(
+            secp,
+            asset.into_tag(),
+            bf.into_inner(),
+        ))
+    }
+
     /// Serialized length, in bytes
     pub fn encoded_length(&self) -> usize {
         match *self {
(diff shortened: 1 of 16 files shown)


src/confidential.rs
```diff
@@ -324,29 +519,155 @@ impl fmt::Display for Nonce {
...
+            p if p == 0x02 || p == 0x03 => {
+                let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
+                Ok(Nonce::Confidential(
+                    PublicKey::from_slice(&bytes).map_err(secp256k1_zkp::Error::Upstream)?,
+                ))
+            }
+            p => return Err(encode::Error::InvalidConfidentialPrefix(p)),
+        }
+    }
+}
```
thomaseizinger: I think we might be able to consolidate these `Decodable` implementations with a macro that is parameterized by:

1. the type you want to construct: `Nonce`
2. the function you need to construct the confidential value: `PublicKey::from_slice`
3. the prefix that identifies the confidential variant: `0x02` and `0x03`

src/confidential.rs
```diff
@@ -324,29 +519,155 @@ impl fmt::Display for Nonce {
...
+            1 => {
+                // ignore prefix when decoding an explicit asset
+                d.consume(1);
+                let explicit = Decodable::consensus_decode(&mut d)?;
+                Ok(Nonce::Explicit(explicit))
+            }
+            p if p == 0x02 || p == 0x03 => {
+                let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
+                Ok(Nonce::Confidential(
+                    PublicKey::from_slice(&bytes).map_err(secp256k1_zkp::Error::Upstream)?,
```
thomaseizinger, line 646: Can you get rid of this `map_err` with a `From` implementation between these two errors?

`?` calls `.from` for you.

src/encode.rs
```diff
@@ -94,7 +94,7 @@ pub trait Encodable {
 /// Data which can be encoded in a consensus-consistent way
 pub trait Decodable: Sized {
     /// Decode an object with a well-defined format
-    fn consensus_decode<D: io::Read>(d: D) -> Result<Self, Error>;
+    fn consensus_decode<D: io::BufRead>(d: D) -> Result<Self, Error>;
```
thomaseizinger, line 117: Maybe mention in the commit message why we are doing this: so we can peek at the prefix and still directly deserialize the commitments.

src/confidential.rs
```diff
@@ -17,251 +17,231 @@
...
+            }
+            1 => {
+                // ignore prefix when decoding an explicit value
+                d.consume(1);
+                let explicit = u64::swap_bytes(Decodable::consensus_decode(&mut d)?);
+                Ok(Value::Explicit(explicit))
+            }
+            p if p == 0x08 || p == 0x09 => {
+                let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
+                Ok(Value::Confidential(PedersenCommitment::from_slice(&bytes)?))
```
thomaseizinger, line 172: You could also implement `Decodable` for `PedersenCommitment` and then simply delegate here. With `BufRead`, we can still peek at the prefix and use it to decide on what to delegate to while retaining the prefix in the buffer so the implementation can use it to decode.

src/confidential.rs
```diff
@@ -272,21 +252,188 @@ impl Asset {
...
+    fn consensus_encode<S: io::Write>(&self, mut s: S) -> Result<usize, encode::Error> {
+        match *self {
+            Asset::Null => 0u8.consensus_encode(s),
+            Asset::Explicit(n) => {
+                1u8.consensus_encode(&mut s)?;
+                Ok(1 + n.consensus_encode(&mut s)?)
+            }
+            Asset::Confidential(generator) => {
+                s.write_all(&generator.serialize())?;
+                Ok(33)
```
thomaseizinger, line 357: Same here, you could implement `Encodable` for `Generator`.

src/confidential.rs
thomaseizinger: Such a macro might be even easier to implement if all the inner types implement `Encodable` / `Decodable`.

src/confidential.rs
```diff
@@ -647,6 +720,68 @@ impl<'de> Deserialize<'de> for Nonce {
     }
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+/// Blinding factor used for asset commitments.
```
thomaseizinger, line 724: These comments are usually on top of the derives.

src/transaction.rs
```diff
@@ -373,10 +375,165 @@ impl Decodable for TxOut {
     }
 }
 
+pub enum TxOutError {
```
thomaseizinger, line 383: Does this implement `std::error::Error`?

src/transaction.rs
```diff
@@ -373,10 +375,165 @@ impl Decodable for TxOut {
...
+            secp,
+            1,
+            value_commitment.commitment().expect("confidential value"),
+            value,
+            out_vbf.0,
+            &message.to_bytes(),
+            address.script_pubkey().as_bytes(),
+            shared_secret,
+            0,
+            52,
```
thomaseizinger, line 433: Should we document the use of these constants better? IIRC, we got them out of the documentation of libwally.

cc @apoelstra

src/transaction.rs
```diff
@@ -671,6 +699,93 @@ impl TxOut {
...
+    /// Scriptpubkey
+    pub script_pubkey: Script,
+    /// There should be no such thing as a nonce for an explicit
+    /// output, but Elements will deserialize such a thing and even
+    /// produce it.
+    pub nonce: Nonce,
+}
+serde_struct_impl!(ExplicitTxOut, asset, value, script_pubkey, nonce);
+
+pub struct ConfidentialTxOut {
```
thomaseizinger, line 717: No derives on this one?

thomaseizinger (commented):
Really good work, thank you @luckysori!

Left some comments, none of them are really pressing though. Take them as you please :)

src/confidential.rs
luckysori: Ooops.

src/confidential.rs
luckysori: Good call!

src/transaction.rs
luckysori: These come from here, yes: https://github.com/ElementsProject/libwally-core/blob/3aa27c1ba1136be5010808c70ff7e8c0f58f66a9/include/wally_elements.h#L142-143

luckysori: They're recommended by the authors of libwally, but it would be better to say why they've been chosen if we are to document them.

apoelstra: The first number, 0, is used to shift the exponent, because CT supports floating-point numbers with an explicit exponent but blind mantissa. Setting this to 0 basically disables the feature, which I think represents a bad privacy/efficiency tradeoff and shouldn't be used.

Setting min-bits to 52 gives you room to represent satoshi values up to 2^52 (45MM BTC, in excess of the total 21MM supply) and was chosen to match the range of a double-width floating point value.

luckysori: > The first number, 0, is used to shift the exponent, because CT supports floating-point numbers with an explicit exponent but blind mantissa. Setting this to 0 basically disables the feature, which I think represents a bad privacy/efficiency tradeoff and shouldn't be used.

What would be a better default value? And should this API expose an optional argument to control it?

apoelstra: Sorry, I worded that poorly. We **should** leave it at 0, and we shouldn't expose any way to change it.

sanket1729 (commented):
Thanks for the PR. Sorry of the delay in reviewing it :( . Can you split the PR into two Prs? The first PR does everything without introducing the new `ExplcitTxOut` and `ConfidentialTxOut` types. The second PR introduces the modeling of different TxOuts. 

I think the first PR would be ready to be and would not require too much discussion while the second one might require some discussion about the merits.

src/confidential.rs
```diff
@@ -17,251 +17,231 @@
...
-    /// Value is committed
-    Confidential(u8, [u8; 32]),
-}
-impl_confidential_commitment!(Value, u64, 0x08, 0x09, u64::swap_bytes);
+#[cfg(feature = "serde")]
+impl Serialize for Value {
+    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        use serde::ser::SerializeSeq;
+
+        let seq_len = if *self == Value::Null { 1 } else { 2 };
```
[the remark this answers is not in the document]
luckysori: Thanks for this comment! This has highlighted that we had introduced some bugs in the serde code which were not showing up because of some missing tests. I'm adding these tests to the split up, less controversial PR which I'll open soon.

luckysori:
Superseded by #77 and #78.

PR #72, opened 2021-02-25
Improve the dynafed::Params fmt::Debug implementation

stevenroose:
The fedpegscript and witness values are Vec<u8> and displayed as arrays
of integers. This implementation changes that to show hex strings for
them instead.

### DIFF
--- a/src/dynafed.rs
+++ b/src/dynafed.rs
@@ -24,7 +24,7 @@ use encode::{self, Encodable, Decodable};
 use Script;
 
 /// Dynamic federations paramaters, as encoded in a block header
-#[derive(Clone, Eq, Hash, PartialEq)]
+#[derive(Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub enum Params {
     /// Null entry, used to signal "no vote" as a proposal
     Null,
--- a/src/dynafed.rs
+++ b/src/dynafed.rs
@@ -75,7 +75,7 @@ impl fmt::Debug for Params {
                 write!(f, "[")?;
                 for (i, e) in self.0.iter().enumerate() {
                     if i != 0 {
-                        write!(f, " ")?;
+                        write!(f, ", ")?;
                     }
                     bitcoin::hashes::hex::format_hex(&e[..], f)?;
                 }
@@ -609,7 +609,7 @@ mod tests {
         };
         assert_eq!(
             to_debug_string(&full),
-            "Full { signblockscript: 0102, signblock_witness_limit: 3, fedpeg_program: 0405, fedpegscript: 0607, extension_space: [0809 0a] }",
+            "Full { signblockscript: 0102, signblock_witness_limit: 3, fedpeg_program: 0405, fedpegscript: 0607, extension_space: [0809, 0a] }",
         );
         let extra_root = full.extra_root();
 
--- a/src/dynafed.rs
+++ b/src/dynafed.rs
@@ -14,17 +14,17 @@
 
 //! Dynamic Federations
 
-use std::io;
+use std::{fmt, io};
 
+use bitcoin;
 use bitcoin::hashes::{Hash, sha256, sha256d};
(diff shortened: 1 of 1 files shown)


src/dynafed.rs
```diff
@@ -538,9 +593,11 @@ mod tests {
             fedpegscript: vec![0x06, 0x07],
             extension_space: vec![vec![0x08, 0x09], vec![0x0a]],
         };
+        println!("full debug: {:?}", full);
```
sanket1729, line 596: Can we do String compare instead of printing? We do not have prints in any of the other tests and it would be helpful for print debugging if we don't have print statements.

src/dynafed.rs
```diff
@@ -24,7 +24,7 @@ use encode::{self, Encodable, Decodable};
 use Script;
 
 /// Dynamic federations paramaters, as encoded in a block header
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Eq, Hash, PartialEq)]
```
sanket1729, line 27: I am curious why did not derive Ord/PartialOrd because I was convinced by your reasoning here(https://github.com/rust-bitcoin/rust-bitcoin/pull/524#issuecomment-734386157)

src/dynafed.rs
stevenroose: donw

src/dynafed.rs
stevenroose: done

stevenroose:
Addressed comments, also rebased.

src/dynafed.rs
```diff
@@ -14,7 +14,7 @@
 
 //! Dynamic Federations
 
-use std::io;
+use std::{fmt, io};
```
apoelstra, line 17: there is a serde-gated import of `fmt` later that you need to remove

src/dynafed.rs
```diff
@@ -56,6 +56,61 @@ pub enum Params {
...
+            }
+        }
+        // ad-hoc struct to fmt in hex
+        struct HexBytesArray<'a>(&'a [Vec<u8>]);
+        impl<'a> fmt::Debug for HexBytesArray<'a> {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(f, "[")?;
+                for (i, e) in self.0.iter().enumerate() {
+                    if i != 0 {
+                        write!(f, " ")?;
```
stevenroose, line 78: Anyone prefer a comma here with the space?

stevenroose:
We could do the same for the serde impl, but that'd be a breaking change.

stevenroose:
Yeah we should have a more common way to do hex. Perhaps even go as far as taking the hex module out of the hashes crate into it's own crate and provide some field tags for serde and for Display/Debug. IIRC, there is a `serde_hex` crate that does this for serde, but not for Display/Debug.

stevenroose:
I pushed the comma and another commit for the serde impl.

apoelstra:
+1 to moving hexing into its own crate. It seemed silly when it was just basic hex encoding/decoding but given that we now have `Display`, `Debug`, `serde`, a couple one-off traits for in-place serialization and deserialization, etc., it's starting to seem more generally useful.

apoelstra (approved):
ack 5ba5bf7828fdf798aa93342bbaf21c45cc39c9bb

Not thrilled about allocating during deserialization but we only do it when parsing the "legacy" serde format, so that's fine.

I'm a little skeptical BTW that this is actually backward-compatible for all serializers. But I wasn't able to think of a reason it wouldn't be beyond serde having hard-to-predict behavior.

RCasatta:
it looks the `github-merge.py` tool doesn't recognize lowercase ack, sorry to bother you @apoelstra  could you please uppercase ack?

apoelstra:
Oh, thanks :)

Sorry, I knew that lowercase acks didn't work, I don't know what I was thinking..

PR #73, opened 2021-03-19
Implement PSET

sanket1729:
This contains an implementation of PSET according to the latest specification in BIP370 and https://github.com/ElementsProject/elements/pull/951.


src/pset/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+        if 0xff_u8 != u8::consensus_decode(&mut d)? {
+            return Err(Error::InvalidSeparator.into());
+        }
+
+        let global: Global = Decodable::consensus_decode(&mut d)?;
+
+        let inputs: Vec<Input> = {
+            let inputs_len = global.n_inputs();
+
+            let mut inputs: Vec<Input> = Vec::with_capacity(inputs_len);
```
apoelstra, line 299: You need to length-check `inputs_len` here (or perhaps above when decoding `global.n_inputs`?). Ditto for the code below regarding outputs.

apoelstra:
Not a huge deal, but can you rebase it to squash the "fix serde bugs" commit into the "implement pset" commit, and change the message to say "implement PSBT2 parts of PSET" or something like that?

src/pset/mod.rs
sanket1729: There is a sanity_check function which is called at the end that does all the checks.

src/pset/map/global.rs
```diff
@@ -80,55 +108,40 @@ pub struct Global {
...
+    /// The version number of this PSET. Must be present.
     pub version: u32,
     /// A global map from extended public keys to the used key fingerprint and
     /// derivation path as defined by BIP 32
     pub xpub: BTreeMap<ExtendedPubKey, KeySource>,
     /// Global proprietary key-value pairs.
-    #[cfg_attr(feature = "serde", serde(with = "::serde_utils::btreemap_as_seq_byte_values"))]
-    pub proprietary: BTreeMap<raw::ProprietaryKey, Vec<u8>>,
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    pub proprietary: GlobalProp,
```
sanket1729, line 118: Not entirely sure if I should be adding these fields to the proprietary field or if I should just add these as normal fields.

sanket1729:
I would really like to get in a `rustfmt` PR if that is possible.

src/pset/mod.rs
sanket1729: So, this saves allocation in case there is a mismatch? Does it matter if we are returning an Error anyways? In any case, I will change it.

sanket1729: I don't follow what you are suggesting? How is possible to check this without the allocation.

apoelstra: By defining a maximum value and then directly comparing it :)

PR #76, opened 2021-03-25
Handle Result returns for engines

stevenroose:


sanket1729:
We should just merge this utACK 3a27b3023fb52aca3bfa90c2977c0a5c75dde2d7 . The same code is repeated in PRs #76 , #70 , #73

sanket1729:
Also, someone with write access would need to configure so disable travis checks in favor of gh actions

apoelstra (approved):
ack 3a27b3023fb52aca3bfa90c2977c0a5c75dde2d7

sanket1729:
I don't know how to avoid this in future, but this exact diff was merged as a part of #77

sanket1729:
So, if we look at the commit history of `master`, it's just one merge commit after another merge commit without contents of this commit

apoelstra:
Heh, oops. I think it's fine, that's just the nature of git.

If we really wanted to avoid it we could insist that every PR be based off of `master` before merging. But I've tried that on a few repos and it's always more hassle than it's worth.

PR #77, opened 2021-04-06
Model Value, Asset and Nonce confidential variants using secp256k1-zkp types

luckysori:
This is the part of PR https://github.com/ElementsProject/rust-elements/pull/70 which should be less controversial.

When splitting up that PR and going through @sanket1729's review I noticed that we had broken the serde implementations of `Value`, `Asset` and `Nonce` (and I found [this](https://github.com/ElementsProject/rust-secp256k1-zkp/pull/10) (our) bug in rust-secp256k1-zkp). 

This also made me realise that, by delegating to said library when implementing serde on our types, we had introduced breaking changes. I've documented the original serde implementations with the tests in https://github.com/ElementsProject/rust-elements/commit/1c53b177c3e56c21836fa3f89bb4b4f2da0689ef, to be able to highlight the changes introduced in https://github.com/ElementsProject/rust-elements/commit/516ac0ee35ed5754ab9e8d76f3cfb60643b51a4b. I made some choices there which could be questionable, so please take a look.

---

Once https://github.com/ElementsProject/rust-secp256k1-zkp/pull/10 is merged, we should remove the patch dependency to `luckysori/rust-secp256k1-zkp` in `Cargo.toml`.

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -17,7 +17,13 @@
 //! Structures representing Pedersen commitments of various types
 //!
 
-use secp256k1_zkp::{self, Generator, PedersenCommitment, PublicKey};
+use bitcoin::hashes::{sha256d, Hash};
+use secp256k1_zkp::{
+    self, compute_adaptive_blinding_factor,
+    ecdh::SharedSecret,
+    rand::{CryptoRng, Rng, RngCore},
+    CommitmentSecrets, Generator, PedersenCommitment, PublicKey, Secp256k1, SecretKey, Signing,
+};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
@@ -38,6 +44,16 @@ pub enum Value {
 }
 
 impl Value {
+    /// Create value commitment.
+    pub fn new_confidential<C: Signing>(
+        secp: &Secp256k1<C>,
+        value: u64,
+        asset: Generator,
+        bf: ValueBlindingFactor,
+    ) -> Self {
+        Value::Confidential(PedersenCommitment::new(secp, value, bf.0, asset))
+    }
+
     /// Serialized length, in bytes
     pub fn encoded_length(&self) -> usize {
         match *self {
@@ -253,6 +269,19 @@ pub enum Asset {
 }
 
 impl Asset {
+    /// Create asset commitment.
+    pub fn new_confidential<C: Signing>(
+        secp: &Secp256k1<C>,
+        asset: AssetId,
+        bf: AssetBlindingFactor,
+    ) -> Self {
+        Asset::Confidential(Generator::new_blinded(
+            secp,
+            asset.into_tag(),
+            bf.into_inner(),
+        ))
+    }
+
     /// Serialized length, in bytes
     pub fn encoded_length(&self) -> usize {
         match *self {
@@ -471,6 +500,50 @@ pub enum Nonce {
 }
 
 impl Nonce {
+    /// Create nonce commitment.
+    pub fn new_confidential<R: RngCore + CryptoRng, C: Signing>(
+        rng: &mut R,
+        secp: &Secp256k1<C>,
+        receiver_blinding_pk: &PublicKey,
+    ) -> (Self, SecretKey) {
+        let sender_sk = SecretKey::new(rng);
+        let sender_pk = PublicKey::from_secret_key(&secp, &sender_sk);
+
+        let shared_secret = Self::make_shared_secret(receiver_blinding_pk, &sender_sk);
+
+        (Nonce::Confidential(sender_pk), shared_secret)
+    }
+
+    /// Calculate the shared secret.
+    pub fn shared_secret(&self, receiver_blinding_sk: &SecretKey) -> Option<SecretKey> {
+        match self {
+            Nonce::Confidential(sender_pk) => {
+                Some(Self::make_shared_secret(&sender_pk, receiver_blinding_sk))
+            }
+            _ => None,
+        }
+    }
+
+    /// Create the shared secret.
+    fn make_shared_secret(pk: &PublicKey, sk: &SecretKey) -> SecretKey {
+        let shared_secret = SharedSecret::new_with_hash(pk, sk, |x, y| {
+            // Yes, what follows is the compressed representation of a Bitcoin public key.
+            // However, this is more by accident then by design, see here: https://github.com/rust-bitcoin/rust-secp256k1/pull/255#issuecomment-744146282
+
+            let mut dh_secret = [0u8; 33];
+            dh_secret[0] = if y.last().unwrap() % 2 == 0 {
+                0x02
+            } else {
+                0x03
+            };
+            dh_secret[1..].copy_from_slice(&x);
+
+            sha256d::Hash::hash(&dh_secret).into_inner().into()
+        });
+
+        SecretKey::from_slice(&shared_secret.as_ref()[..32]).expect("always has exactly 32 bytes")
+    }
+
     /// Serialized length, in bytes
     pub fn encoded_length(&self) -> usize {
         match *self {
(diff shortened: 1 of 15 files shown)


src/confidential.rs
```diff
@@ -419,22 +741,29 @@ mod tests {
...
-                Token::Tuple { len: 32 },
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::U8(1), Token::U8(1), Token::U8(1), Token::U8(1),
-                Token::TupleEnd,
```
thomaseizinger: So this used to serialize as a tuple? Very interesting! I guess for a serializer impl like JSON, it doesn't make a difference because JSON doesn't have support for tuples?

src/encode.rs
```diff
@@ -53,9 +58,17 @@ impl fmt::Display for Error {
             Error::OversizedVectorAllocation {
                 requested: ref r,
                 max: ref m,
-            } => write!(f, "oversized vector allocation: requested {}, maximum {}", r, m),
+            } => write!(
+                f,
+                "oversized vector allocation: requested {}, maximum {}",
+                r, m
+            ),
```
thomaseizinger, line 68: Nit: This is just a format change, not sure if it is wanted / tolerated?

src/transaction.rs
```diff
@@ -1661,7 +1660,7 @@ mod tests {
 
         // Output with pushes that are e.g. OP_1 are nulldata but not pegouts
         let output: TxOut = hex_deserialize!("\
-            0a2d3634393536d9a2d0aaba3823f442fb24363831fdfd0101010101010101010\
+            0a319c0000000000d3d3d3d3d3d3d3d3d3d3d3d3fdfdfd0101010101010101010\
```
thomaseizinger, line 1870: Do you remember why this changed?

src/transaction.rs
```diff
@@ -373,10 +375,165 @@ impl Decodable for TxOut {
     }
 }
 
+pub enum TxOutError {
+    NoBlindingKeyInAddress,
+    Upstream(secp256k1_zkp::Error),
+}
```
thomaseizinger: This should implement `std::error::Error`! :)

src/transaction.rs
```diff
@@ -376,13 +376,35 @@ impl Decodable for TxOut {
...
 /// Errors encountered when constructing confidential transaction outputs.
+#[derive(Debug, Clone, Copy)]
 pub enum ConfidentialTxOutError {
     /// The address provided does not have a blinding key.
     NoBlindingKeyInAddress,
     /// Error originated in `secp256k1_zkp`.
     Upstream(secp256k1_zkp::Error),
 }
 
+impl fmt::Display for ConfidentialTxOutError {
```
thomaseizinger, line 387: Can we squash this into the one where it is introduced?

src/transaction.rs
```diff
@@ -404,6 +404,11 @@ impl std::error::Error for ConfidentialTxOutError {
     }
 }
 
+impl From<secp256k1_zkp::Error> for ConfidentialTxOutError {
```
thomaseizinger, line 407: Also a candidate for squashing I'd say.

src/hash_types.rs
```diff
@@ -55,9 +52,3 @@ impl_hashencode!(Wtxid);
 impl_hashencode!(SigHash);
 impl_hashencode!(BlockHash);
 impl_hashencode!(TxMerkleNode);
-
-impl ThirtyTwoByteHash for SigHash {
```
thomaseizinger: If this is deleted here, we should just drop the previous commit.

src/script.rs
```diff
@@ -1158,8 +1158,6 @@ mod test {
         assert!(script_2 < script_3);
         assert!(script_3 < script_4);
 
-        assert_eq!(script_1, script_1);
```
thomaseizinger: :grin:

thomaseizinger (commented):
Some nits, mainly concerning patch history :)

Thanks for pushing the upstream forward!

src/transaction.rs
sanket1729: The test was incorrect. The asset is not a correct curve point

src/script.rs
```diff
@@ -1147,7 +1147,7 @@ mod test {
         assert_eq!(v_nonmin_alt.unwrap(), slop_v_nonmin_alt.unwrap());
     }
 
-	#[test]
+        #[test]
```
sanket1729: indent

src/script.rs
```diff
@@ -1158,8 +1158,7 @@ mod test {
         assert!(script_2 < script_3);
         assert!(script_3 < script_4);
 
-        assert!(script_1 <= script_1);
```
sanket1729, line 1161: These lines should not be deleted

sanket1729 (commented):
Overall, I think this looks good, left some minor code style nits. It would be great to also incorporate suggestions which Thomas Eizenger suggested.

I am not sure about the `serde` breaking change. I will let @apoelstra comment on it.

src/hash_types.rs
luckysori: I must have messed something up, because this change was desired. I'll fix it.

src/script.rs
luckysori: This is bugging me because locally I don't have this change D:

luckysori:
Squashed and I tried to deal with all the nits. Maybe someone else can deal with the silly indent in `script.rs` line 1150, because it's not showing up on my machine :shrug:

src/script.rs
luckysori: Just curious, what's this supposed to catch? Doesn't the compiler enforce that `script_1` is equal to `script_1`? I've replaced it with `assert_eq`, but maybe you actually wanted the original assertions. Let me know.

src/script.rs
thomaseizinger: Make sure you are diffing against current master. Maybe another rebase helps!

src/script.rs
sanket1729: I think the test is script is for `script_ord` which is supposed to check basic comparison orderings like `script_1 >= script_1`. Even if it tests something that is trivial, it should not be removed as a part of this PR.

luckysori: I've reverted it back to the original state.

apoelstra:
Sorry for the long delay in reviewing.

concept ACK everything, including the breaking serde changes. (I'm not thrilled about using a sequence of (tag, bytes) like this, but I like my original scheme even less .... would prefer if we serialized as a flat sequence of bytes or a hex string, similar to what `PublicKey` et al do).

There are two problems blocking this now:

1. The unit tests fail with `cargo test --features="serde-feature"` as of 1c53b177c3e56c21836fa3f89bb4b4f2da0689ef
2. The cargo patch is still in place as of 285f5656 ... and this doesn't compile anymore since the branch has been deleted!

Unsure why CI is not catching the unit test failure :( does anyone else see this?

sanket1729:
@apoelstra , I think both of those are related to deleted branch. I have the branch saved locally, so all commits in the PR compile and test with `serde-feature` correctly. 

The branch was probably not deleted when the CI ran, so the CI did not detect the failure

sanket1729:
Looks like the branch was merged into rust-secp-zkp master, that is why it was deleted.

sanket1729:
rust-secp-zkp `0.2.1` has been released. So, we don't need any patch dependencies :)

apoelstra:
Confirmed that a modified version of this PR with the `[patch]` removed now passes (unit tests pass on every commit)

apoelstra:
Sorta-ack 7a248e0e7f8b464890e3137b62011524e394a237

This PR looks great! Thanks so much for this work, especially doing the fiddly computations for last vs not-last `TxOut`s. If you can remove the `[patch]` from Cargo.toml I'm happy to merge this.

luckysori:
@apoelstra 

> Sorta-ack [7a248e0](https://github.com/ElementsProject/rust-elements/commit/7a248e0e7f8b464890e3137b62011524e394a237)

I actually noticed that I had set `RANGEPROOF_MIN_VALUE` to 0 instead of 1. Maybe it's not a problem, but I've changed it back to 1 (see 38e89fb).

> This PR looks great! Thanks so much for this work, especially doing the fiddly computations for last vs not-last `TxOut`s. If you can remove the `[patch]` from Cargo.toml I'm happy to merge this.

Thank you! I've now removed the patch dependency from all commits.

PR #78, opened 2021-04-06
Allow unblinding of confidential TxOuts

luckysori:
For this we allow the conversion of `TxOut` into `ExplicitTxOut` and `ConfidentialTxOut`. We then implement `unblind` on the latter.

### DIFF
--- a/src/blind.rs
+++ b/src/blind.rs
@@ -17,7 +17,7 @@
 
 use std::{self, fmt};
 
-use secp256k1_zkp::{self, PedersenCommitment, Tag, Tweak, ZERO_TWEAK, rand::{CryptoRng, RngCore}};
+use secp256k1_zkp::{self, PedersenCommitment, SecretKey, Tag, Tweak, Verification, ZERO_TWEAK, rand::{CryptoRng, RngCore}};
 use secp256k1_zkp::{Generator, RangeProof, Secp256k1, Signing, SurjectionProof};
 
 use AddressParams;
@@ -27,6 +27,7 @@ use {Address, AssetId, Transaction, TxOut, TxOutWitness,
     ValueBlindingFactor
 }};
 
+use hashes;
 
 /// Transaction Output related errors
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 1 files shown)


src/transaction.rs
```diff
@@ -572,6 +575,48 @@ impl TxOut {
...
+    /// Convert into a confidential output if possible.
+    pub fn to_confidential(&self) -> Option<ConfidentialTxOut> {
+        Some(ConfidentialTxOut {
+            asset: self.asset.commitment()?,
+            value: self.value.commitment()?,
+            nonce: self.nonce,
+            script_pubkey: self.script_pubkey.clone(),
+            witness: self.witness.clone(),
+        })
+    }
```
thomaseizinger, line 618: In case there is controversy about these, we could also add the `unblind` API directly on `TxOut` and simply extend the `Error` type with a variant `NotConfidential`.

apoelstra: Yeah, I would prefer this approach. In principle you can have txouts that are "half blinded" (explicit asset confidential value, or vice versa) and I'd like to avoid excluding these in the type system.

luckysori: @apoelstra I've adapted the PR based on this thread and the changes that have been introduced since I opened it.

src/blind.rs
```diff
@@ -447,10 +448,106 @@ impl TxOut {
...
         Ok((txout, out_abf, out_vbf))
     }
+
+    /// Unblinds a transaction output, if it is confidential.
+    ///
+    /// It returns the secret elements of the value and asset Pedersen commitments.
+    pub fn unblind<C: Verification>(
+        &self,
+        secp: &Secp256k1<C>,
+        blinding_key: SecretKey,
```
apoelstra, line 458: I think now (since #82) this should be a `ValueBlindingFactor`

luckysori: I think value/asset blinding factors are different to the blinding key used in combination with the nonce to unblind the output. In fact, the value blinding factor of the output is returned as part of unblinding it.

Perhaps you are referring to the fact that blinding factors are now modelled as `Tweak`s in `secp256k1-zkp`. I'm not sure if the "blinding key" used to unblind the output deserves the same treatment, or if we can still model it as a secret key.

apoelstra: You're right, I was confused by our poor terminology here.

sanket1729:
utACK 33c58da

PR #80, opened 2021-04-19
One shot cargo fmt

sanket1729:
I think we should do it earlier when the diff is smaller and the project is smaller in scope. Otherwise, it becomes a big task to do it file by file as with rust-bitcoin. So far, we have had good experiences with rust-miniscript. Most rust-bitcoin developers are also in favor of using it, and I think we should get it done when the diff is relatively small and very easy to review/reproduce.

 
Curious what do @apoelstra, @stevenroose, and others think about this.


src/transaction.rs
```diff
@@ -1480,34 +1542,33 @@ mod tests {
...
                 extra_data: vec![
                     &[
-                        0x02,
-                        0xe2, 0x5e, 0x58, 0x2a, 0xc1, 0xad, 0xc6, 0x9f,
-                        0x16, 0x8a, 0xa7, 0xdb, 0xf0, 0xa9, 0x73, 0x41,
-                        0x42, 0x1e, 0x10, 0xb2, 0x2c, 0x65, 0x99, 0x27,
-                        0xde, 0x24, 0xfd, 0xac, 0x6e, 0x9f, 0x1f, 0xae,
+                        0x02, 0xe2, 0x5e, 0x58, 0x2a, 0xc1, 0xad, 0xc6, 0x9f, 0x16, 0x8a, 0xa7,
+                        0xdb, 0xf0, 0xa9, 0x73, 0x41, 0x42, 0x1e, 0x10, 0xb2, 0x2c, 0x65, 0x99,
+                        0x27, 0xde, 0x24, 0xfd, 0xac, 0x6e, 0x9f, 0x1f, 0xae,
```
apoelstra, line 1555: this kinda thing really really bugs me about rustfmt

apoelstra:
Yeah, I think I can live with this

src/transaction.rs
RCasatta: #[rustfmt::skip] on this bytes vec fields?

sanket1729:
@RCasatta I am still seeking concept ACKs on this. What are your opinions on cargo fmt?

RCasatta:
> @RCasatta I am still seeking concept ACKs on this. What are your opinions on cargo fmt?

cargo fmt is a strong concept ACK for me.

Sometimes it's not as perfect as manual formatting, like not having bytes vec in octets or spanning a method chain in million lines. However, the benefit of having an automatic check in CI and less formatting nits turnaround outweighs the downsides.

Also, I personally find it useful to write code without thinking about formatting cause I know you can fix it easily and without a doubt.

contrib/test.sh
```diff
@@ -41,3 +41,11 @@ then
         cargo test --features integration
     )
 fi
+# Lint if told to
+if [ "$DO_LINT" = true ]
+then
+    (
+        rustup component add rustfmt
```
RCasatta, line 48: it looks it is already installed, CI prints:
`info: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is up to date`

sanket1729:
I plan to go through the diff and add another commit adding `rustfmt::skip`.

sanket1729:
Based on 
Yeah, I think I can live with this - @apoelstra 
cargo fmt is a strong concept ACK for me - @RCasatta 

and a minor complaint
I'm generally not a fan of formatters. ...won't veto anything ofc - @stevenroose 

I will open this for review after adding `rustfmt::skip` to oddly formatted things.

delta1:
concept ACK - I don't ever want to think about formatting

sanket1729:
Reviewing this with a refreshed ACK from @delta1 and the comments posted before. https://github.com/ElementsProject/rust-elements/pull/80#issuecomment-995969709

apoelstra:
Closing this. The next time we upgrade our rust-bitcoin crates I think we should revisit formatting.

PR #81, opened 2021-05-04
Update to new types from rust-secp-zkp

sanket1729:
First part of #73 . Implementation of pset


src/pset/map/global.rs
```diff
@@ -236,13 +244,23 @@ impl Map for Global {
 
         // Serializing version only for non-default value; otherwise test vectors fail
         if self.version > 0 {
+            let ver = self.version; //hack to use macro
+            impl_pset_get_pair!(
+                rv.push_mandatory(ver as <PSET_GLOBAL_VERSION, _>)
+            )
+        }
```
apoelstra, line 251: Now that this is a mandatory field, we should serialize it unconditionally I think.

src/pset/map/global.rs
```diff
@@ -426,14 +473,21 @@ impl Decodable for Global {
...
-            let mut rv: Global = Global::from_unsigned_tx(tx)?;
-            rv.version = version.unwrap_or(0);
-            rv.xpub = xpub_map;
-            rv.unknown = unknowns;
-            Ok(rv)
-        } else {
-            Err(Error::MustHaveUnsignedTx.into())
-        }
+        // Mandatory fields
+        let version = version.ok_or(Error::MissingPsetVersion)?;
```
apoelstra, line 477: We should require the version to be 2, not just that it be present.

src/pset/mod.rs
```diff
@@ -180,19 +178,19 @@ impl PartiallySignedTransaction {
...
         let mut outputs = vec![];
 
-        for psetin in self.inputs.into_iter() {
+        for psetin in self.inputs.iter() {
             let txin = TxIn {
                 previous_output: OutPoint::new(psetin.previous_txid, psetin.previous_output_index),
                 is_pegin: psetin.previous_output_index & (1 << 30) != 0,
                 has_issuance: psetin.previous_output_index & (1 << 31) != 0,
-                script_sig: psetin.final_script_sig.unwrap_or(Script::new()),
+                script_sig: psetin.final_script_sig.clone().unwrap_or(Script::new()),
```
apoelstra, line 193: I know this is the same as the old code, but FYI we could use `unwrap_or_default` here (and a few other places)

apoelstra:
This looks great!!

Can you add a new `deserialize_pset` fuzztest target with

```
fn do_test(data: &[u8]) {
    let psbt: Result<elements::pset::PartiallySignedTransaction, _> = elements::encode::deserialize(data);
    match psbt {
        Err(_) => {}, 
        Ok(psbt) => {
            let ser = elements::encode::serialize(&psbt);
            let deser: elements::pset::PartiallySignedTransaction = elements::encode::deserialize(&ser).unwrap();
            // Since the fuzz data could order psbt fields differently, we compare to our deser/ser instead of data
            assert_eq!(ser, elements::encode::serialize(&deser));
        }   
    }   
}
```

? This triggers right now if you try to parse a PSET with version 0 but it should be fixed by requiring the version to be 2.

src/pset/map/global.rs
sanket1729: It's only a mandatory field in v2

src/pset/map/global.rs
sanket1729: Sure, will fix this. I was thinking about future versions.

sanket1729:
I think this is expected behavior based on spec for v0. But as you say fixing this field to 2 solves all the issues.

sanket1729:
@apoelstra made the requested changes and added the fuzz tests.

src/pset/map/global.rs
apoelstra: But there are no other versions of PSET.

apoelstra (approved):
ack a8397ef481aa0a348cffaf3d19b918f6263f93f3

PR #83, opened 2021-05-25
Pset pr 3

sanket1729:


src/blind.rs
```diff
@@ -137,12 +137,20 @@ impl std::error::Error for VerificationError {}
 /// Errors encountered when constructing confidential transaction outputs.
 #[derive(Debug, Clone, Copy)]
 pub enum ConfidentialTxOutError {
```
apoelstra, line 139: FYI this is a breaking change, so this PR should get in before the release one.

sanket1729: Sure

src/pset/mod.rs
```diff
@@ -248,13 +253,296 @@ impl PartiallySignedTransaction {
...
+            Vec<(u64, AssetBlindingFactor, ValueBlindingFactor)>,
+            Vec<usize>,
+        ),
+        PsetBlindError,
+    > {
+        if inp_txout_sec.len() != self.inputs.len() {
+            return Err(PsetBlindError::InputTxOutSecretLen);
+        }
+
+        let mut blind_out_indices = Vec::new();
```
apoelstra, line 272: nit: can use `Vec::with_capacity(self.outputs.len())`

src/pset/mod.rs
```diff
@@ -248,13 +253,296 @@ impl PartiallySignedTransaction {
...
+            .iter()
+            .filter(|o| o.is_some())
+            .map(|o| o.unwrap())
+            .map(|o| (o.value, o.asset_bf, o.value_bf))
+            .collect::<Vec<_>>();
+
+        if blind_out_indices.is_empty() {
+            // Atleast one output should be blinded in this function
+            return Err(PsetBlindError::AtleastOneOutputBlind);
+        }
```
apoelstra, line 304: I'm a little confused here -- I think it's perfectly acceptable if no outputs get blinded, right?

src/pset/mod.rs
```diff
@@ -248,13 +253,296 @@ impl PartiallySignedTransaction {
...
+                self.outputs[i].ecdh_pubkey = txout.nonce.commitment().map(|pk| bitcoin::PublicKey{
+                    key: pk,
+                    compressed: true
+                });
+            }
+            // return blinding factors used
+            ret.push((abf, vbf));
+        }
+
+        let (value, abf, vbf) = *out_secrets.last().unwrap();
```
apoelstra, line 372: Here we can use `.pop()` rather than `.last()`, and then below we don't need to `len() - 1`.

Then to avoid issues in the case that there are no outputs to be blinded, we can do `unwrap_or(0, ValueBlindingFactor::zero(), ValueBlindingFactor::zero())`.

apoelstra: err `unwrap_or((0, ZERO_TWEAK, ZERO_TWEAK))`

src/pset/mod.rs
```diff
@@ -248,13 +253,296 @@ impl PartiallySignedTransaction {
...
+        &mut self,
+        rng: &mut R,
+        secp: &secp256k1_zkp::Secp256k1<C>,
+        inp_txout_sec: &[Option<&TxOutSecrets>],
+    ) -> Result<(), PsetBlindError> {
+        let (_inp_secrets, mut outs_to_blind) = self.blind_checks(inp_txout_sec)?;
+
+        // If there are more than 1 outs to blind
+        // blind all outs but the last one
+        let last_out_index = outs_to_blind.pop().unwrap();
```
apoelstra, line 429: ...and if `blind_checks` didn't error out on "nothing to blind", this `unwrap` needs to be an error return

src/pset/mod.rs
```diff
@@ -248,13 +253,296 @@ impl PartiallySignedTransaction {
...
+                // Valid secret key
+                let scalar =
+                    SecretKey::from_slice(value_diff.as_ref()).expect("Valid non-zero tweak");
+                final_blind
+                    .add_assign(scalar.as_ref())
+                    .expect("Valid tweak");
+            }
+        }
+        // The last vbf
+        let out_vbf = ValueBlindingFactor::from_slice(final_blind.as_ref()).expect("32 byte");
```
apoelstra, line 475: Can we implement `add_assign` directly on `ValueBlindingFactor`, and/or implement a conversion funcion from `SecretKey`? It would eliminate a lot of this awkwardness.

src/pset/mod.rs
sanket1729: It is bad to assign worst-case capacity when we might not end up using it? Suppose we have a tx with 4000 outputs and only 1 to blind, does this allocate memory for 4000 capacity?

src/pset/mod.rs
sanket1729: I have added a early return if outs_to_blind = 0, so can safely unwrap() here.

src/pset/mod.rs
sanket1729: See comment above

sanket1729:
Also reorganized some of the commits to be more atomic.

src/pset/mod.rs
apoelstra: It is more idiomatic to eliminate the `unwrap`s, and if we're early-returning rather than panicking, to do this directly rather than having a separate branch

It is also more idiomatic to just use `pop` rather than using `last` and then using extra arithmetic to ignore the last item.

apoelstra (approved):
ack 69d73f4900ebee656e4fe0577a3cef65ad570af7

I'm still not thrilled with the `unwrap` but it's not a big deal, it's clearly safe.

PR #88, opened 2021-08-05
Export blind errors

LeoComandini:
BlindError, UnblindError and PsetBlindError are used by public
functions but are not exported.

However I am not sure those were meant to be exposed.
Perhaps they should have been variants of already exported errors.


sanket1729 (approved):
utACK e4392910e63773a4579377d3e2ae5f7f1bb0b36f

PR #90, opened 2021-08-19
pset: set last vbf correctly with one blinded output

LeoComandini:
In case we have a single blinded output and at least one blinded
input, the transaction can be blinded successfully. However
pset::blind_last was ignoring the inputs contributions when
computing the last value blinding factor, causing a
BalanceCheckFailed error. This commit adds a test and a fix.

Not sure if the test is in the right place, I took most of its parts from the examples.


apoelstra (approved):
ACK 707276676084ac37626afcebf374bd31b646a900

PR #92, opened 2021-08-27
OutPoint: implement null and is_null

LeoComandini:
Same as in rust-bitcoin.


apoelstra (approved):
ACK 3457a760b53608c086c6d62ef7f3e4acb28806d5

PR #99, opened 2021-11-09
Add dynafed getter helpers on BlockHeader

stevenroose:
It turns out really painful to get the current parameters from a header.


apoelstra (approved):
ACK e8c848f5636c7dc8c3294885d366caea4c3a1dd1

PR #100, opened 2021-11-10
Update pset spec

sanket1729:
This updates the spec to the latest pset spec. Includes support for blinding, but does not include support for blind value/asset proofs.


src/pset/map/output.rs
```diff
@@ -75,12 +89,16 @@ pub struct Output {
     /// corresponding master key fingerprints and derivation paths.
     #[cfg_attr(feature = "serde", serde(with = "::serde_utils::btreemap_as_seq"))]
     pub bip32_derivation: BTreeMap<PublicKey, KeySource>,
-    /// (PSET2) The amount of the output
-    pub amount: confidential::Value,
+    /// (PSET2) The explicit amount of the output
+    pub amount: Option<u64>,
+    /// (PSET2) The out amount commitment
+    pub amount_comm: Option<secp256k1_zkp::PedersenCommitment>,
```
apoelstra, line 94: In 9840052067bd91bf09cf65001f307f95a47abca1:

I think you should drop `PSET2` from all of these. There are no other versions of PSET.

src/pset/map/output.rs
```diff
@@ -117,13 +166,46 @@ impl Output{
         rv
     }
 
+    /// Create a output from unblinded txout
+    /// This is to be used when the txout is not blinded. This sets
+    /// the blinding key from the txout nonce
```
apoelstra, line 170: In what contexts is this function actually used? Its purpose seems to be to deal with a transaction which has been partially created using the old Elements rawtransactions API, in which case it should have a more obscure name (if it should be used at all).

apoelstra:
ACK 9f33f607d3873fe37bbbf4376c5ead25abad11eb except for the nits that I mentioned

In followup work (or an additional commit) we need to update `unique_id()` to correctly use the explicit value data, if available.

src/pset/map/output.rs
sanket1729: It is how it is currently implemented in el`ementsd`. There is some RPC that creates a psbt from raw transaction and we need to mimic it's behaviour.

sanket1729:
Removed PSET2 references.

> In followup work (or an additional commit) we need to update unique_id() to correctly use the explicit value data, if available.

Is this not how it is currently implemented?

RCasatta:
I tested this branch against the final pset created from https://github.com/ElementsProject/elements/blob/master/contrib/assets_tutorial/pset_swap_tutorial.py (the one included in 225896a74b8c39feec40677c0a736374cc1cef3c)
And there are errors in parsing RangeProofs and SurjectionProofs -> "data not consumed entirely when explicitly deserializing" 

The reason appears to be that for RangeProof the `raw_value` in `impl_pset_prop_insert_pair!` is missing the prepended varint indicating the length of the vector.

Even worst, it looks the varint is missing only in this pset, while in the existing test cases it appears to be there.

An ugly workaround to pass old test cases and the new one is ac56a8e4c2a0d839719812e92582cada6607ee64, I am not sure how to fix properly

src/pset/map/output.rs
```diff
@@ -117,13 +165,46 @@ impl Output{
...
+    pub fn from_txout(txout: TxOut) -> Self {
```
LeoComandini, line 171: in `PartiallySignedTransaction::from_tx`
```
let outputs = tx.output.into_iter().map(Output::from_txout).collect();
```
Should be changed to use `from_blinded_txout` if the output is blinded.

src/pset/map/output.rs
```diff
@@ -117,13 +165,46 @@ impl Output{
...
+                (None, None) => confidential::Asset::Null,
+            },
+            value: match (self.amount_comm, self.amount) {
+                (Some(comm), _) => confidential::Value::Confidential(comm),
+                (None, Some(x)) => confidential::Value::Explicit(x),
+                (None, None) => confidential::Value::Null,
+            },
             nonce: self
-                .ecdh_pubkey
+                .blinding_key
```
LeoComandini, line 207: This should be `ecdh_pubkey` if the output is blinded and `blinding_key` if the output is unblinded.

RCasatta:
expanding on https://github.com/ElementsProject/rust-elements/pull/100#issuecomment-980089032

I think rust-elements is serializing variable length fields, like `RangeProof` and `SurjectionProof`, in PSET pairs wrong. The `raw_value` in `Pair` is already encoded with the length at the `Pair` level, but currently calling `encode::serialize` on `RangeProof` and `SurjectionProof` will add another varint in the `raw_value` because they are `Vec<u8>`, this inner varint is not there in PSET created from elements 0.21.

Since we cannot change the consensus_encode on RangeProof and Surjection proof, in another WIP commit ad3c86f2b3013414f640a0e0bd473f9e02bf2319 I am adding a macro case for these fields to call directly `RangeProof::serialize` and `SurjectionProof::serialize` to skip the initial varint.

This breaks some test cases in `test_pset` but I thought that maybe they are created with rust-elements itself and test vectors are wrong.
Moreover, even the round-trip in `pset_from_elements` is still-failing, but the PSET serialization length is now the same (wasn't the same before because of the added varint) and tests from @LeoComandini accept the generated PSET in elements-core. (Maybe the hex comparison could break because of field order?)

src/pset/map/output.rs
RCasatta: something along the line of ecfe679b2db625e628515aa95a0c54d1f750d6f5. 
Maybe we can also remove `from_blinded_txout` and do the `if output.is_partially_blinded()` inside `from_txout`

src/pset/map/output.rs
RCasatta: something along the line of ecfe679

apoelstra:
This sounds good to me. I'm fine with breaking test vectors in this library (which likely were created with the library itself), since Elements is the reference (and Elements seems to be doing something reasonable)

RCasatta:
I'll clean up my branch and prepare a PR over this one

sanket1729:
> Even worst, it looks the varint is missing only in this pset, while in the existing test cases it appears to be there.

rust-elements had an implementation of PSET even before elements. So, the test vectors were created using this library. I got bust with other projects that I did not cross-test with elements after it was implemented. 

> I'll clean up my branch and prepare a PR over this one

Thanks for doing this.

sanket1729:
@RCasatta Are you sure that the decoding bug is related to `Rangeproof` and `SurjectionProof`. There is an unfixed bug in elements that might be causing the problem, See: https://github.com/ElementsProject/elements/issues/1065 .

AFAIR, I tested that rangeproofs and surjectoinproofs work correctly

sanket1729:
Also related: https://github.com/ElementsProject/rust-elements/issues/101

RCasatta:
> @RCasatta Are you sure that the decoding bug is related to `Rangeproof` and `SurjectionProof`. There is an unfixed bug in elements that might be causing the problem, See: [ElementsProject/elements#1065](https://github.com/ElementsProject/elements/issues/1065) .
> 
> AFAIR, I tested that rangeproofs and surjectoinproofs work correctly

Not sure about this other bugs, but I think `RangeProof` and `SurjectionProof` have the problem

to test deserialization failure, start from this branch then:
* cherry-pick pset test ef6712b7da7f0ba89372c7f5f0c52986d45d8347 (you can obtain the hex by printing the complete var https://github.com/ElementsProject/elements/blob/7d83cc0089345a0646834986c56e58543fd5ee07/contrib/assets_tutorial/pset_swap_tutorial.py#L340 and converting to hex)
* apply the diff (old style debug with println )
```
diff --git a/src/pset/map/input.rs b/src/pset/map/input.rs
index 7a5496d..4636640 100644
--- a/src/pset/map/input.rs
+++ b/src/pset/map/input.rs
@@ -434,7 +434,9 @@ impl Map for Input {
                             impl_pset_prop_insert_pair!(self.issuance_asset_entropy <= <raw_key: _> | <raw_value : [u8;32]>)
                         }
                         PSBT_ELEMENTS_IN_UTXO_RANGEPROOF => {
-                            impl_pset_prop_insert_pair!(self.in_utxo_rangeproof <= <raw_key: _> | <raw_value : RangeProof>)
+                            println!("1");
+                            impl_pset_prop_insert_pair!(self.in_utxo_rangeproof <= <raw_key: _> | <raw_value : RangeProof>);
+                            println!("2");
                         }

```
* run `cargo test -- pset_from_elements --nocapture`

See 1 is printed before the error but 2 it's not

src/pset/map/output.rs
sanket1729: Agreed in principle. Though note that this will create a diversion from how elements `createpsbt` and `converttopsbt` work. 
```
if (txout.nNonce.IsCommitment()) {
    output.m_blinding_pubkey.Set(txout.nNonce.vchCommitment.begin(), txout.nNonce.vchCommitment.end());
}
```

The above RPCs don't check any status of blinding of the transaction and assume the transaction to be unblinded. I am okay(and in support of) with changing this here but highlighting that this is different from elements core RPC.

sanket1729:
@RCasatta, how about a707fe4? I updated the is_partially_confidential() to remove the nonce field check from it.
If the txout is [TxOut::is_partially_blinded], then nonce of the txout is treated as ecdh_pubkey, otherwise the nonce of the txout is assumed to be receiver blinding key.

src/pset/map/output.rs
sanket1729: Did something similar in the last commit

RCasatta:
ACK dc350a0b8664cdd5b051443810a6a1f99a6173da

LeoComandini:
ACK dc350a0b8664cdd5b051443810a6a1f99a6173da, code review, tested with GDK and against Elements 0.21

PR #104, opened 2021-11-24
Remove pub accessor to PSET inputs and outputs 

### LINKED ISSUE #103: Remove pub from PSET inputs and outputs
Having inputs and outputs pub encourage a library user to write code like this

pset.inputs.push(Input::default());
Which will results in wrong serialization because self.global.tx_data.output_count is not updated accordingly like it's done in add_input().

Obviously, accessor/mutator will be needed to operate on inputs/outputs but those will preserve PSET internal invariants

RCasatta:
This removes `pub` accessor to PSET inputs and outputs fields, but provides accessor/mutator functions that preserve global count

close #103

### DIFF
--- a/src/pset/mod.rs
+++ b/src/pset/mod.rs
@@ -53,10 +53,10 @@ pub struct PartiallySignedTransaction {
     pub global: Global,
     /// The corresponding key-value map for each input in the unsigned
     /// transaction.
-    pub inputs: Vec<Input>,
+    inputs: Vec<Input>,
     /// The corresponding key-value map for each output in the unsigned
     /// transaction.
-    pub outputs: Vec<Output>,
+    outputs: Vec<Output>,
 }
 
 impl PartiallySignedTransaction {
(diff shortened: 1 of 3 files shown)


src/pset/mod.rs
```diff
@@ -91,13 +91,47 @@ impl PartiallySignedTransaction {
...
+    pub fn inputs(&self) -> &[Input] {
+        &self.inputs
+    }
+    /// Mutable accessor to inputs
+    pub fn inputs_mut(&mut self) -> &mut [Input] {
+        &mut self.inputs
+    }
+    /// Remove the input at `index` and return it if any, otherwise returns None
+    /// This also updates the pset global input count
+    pub fn remove_input(&mut self, index:usize) -> Option<Input> {
```
stevenroose, line 104: Could you put newlines between the methods and a space after this colon? And the similar colon for outputs :)

stevenroose (changes requested):

stevenroose:
Is this change equivalent to a PSBT in rust-bitcoin API change?

src/pset/mod.rs
RCasatta: updated

RCasatta:
> Is this change equivalent to a PSBT in rust-bitcoin API change?

Not really, something similar in PSBT "v0" is more complicated I think because you have to mantain coherence between tx and psbt inputs/outputs.
In PSBTv2 I am not aware of open PRs

@LeoComandini just questioned me about the existence of global_count at all and I don't know why it exists, now I noted it's not in the spec, so maybe we just have to remove the counter and we can keep the `pub`

sanket1729:
> now I noted it's not in the spec, so maybe we just have to remove the counter and we can keep the pub

PSET is based on PSBTv2 (BIP 370). It includes all fields in psbtv2, plus the additional fields. [Link](https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki)

RCasatta:
> > now I noted it's not in the spec, so maybe we just have to remove the counter and we can keep the pub
> 
> PSET is based on PSBTv2 (BIP 370). It includes all fields in psbtv2, plus the additional fields. [Link](https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki)

Sorry, I checked only https://github.com/ElementsProject/elements/blob/master/doc/pset.mediawiki

I thought that we can remove `input_count` and `output_count` from `TxData` and use just `inputs.len()` and `outputs.len()` during serializaton, but the problem in that is checking they are consistent during deserialization...

So yeah I think this MR is needed

stevenroose:
In what scenario would you want to remove inputs or outputs from the PSET, but not from the underlying tx? I thought usually first a TX was created and then a PSBT wrapper made and from then on the tx wouldn't change anymore?

RCasatta:
> In what scenario would you want to remove inputs or outputs from the PSET, but not from the underlying tx? I thought usually first a TX was created and then a PSBT wrapper made and from then on the tx wouldn't change anymore?

With PSBT "v0" is like you describe but PSBTv2 is designed to allow changes (there isn't anymore an unsigned tx...), see https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki

sanket1729:
The cleaner solution involves removing the `n_input` and `n_output` from Global field and making sure that serialization and deserialization are implemented accordingly. But because of the current architectural separation of Global, Input, and Output, I think it would be difficult to implement serialization for it.

sanket1729:
Or, we can make `n_inputs` and `n_outputs` private. That should be the simplest way to fix this while maintaining convenience?

RCasatta:
So I think we agree cleaner solutions removing inputs and outputs counters would require a bigger refactor, so I think we can go with this MR for now

sanket1729 (approved):
core review ACK b7c0831b70bbe055d553bcad6bc057f803967d9a

RCasatta:
Sorry, I merged this PR wrong, merging the commit b7c0831b70bbe055d553bcad6bc057f803967d9a in the merge commit itself. Luckily the commit was mine and I didn't de-authored anyone, so I think it's ugly but fine. Let me know if there are issues

PR #114, opened 2022-02-08
Fix BlockHeader roundtrip bug on serde binary encoding

### LINKED ISSUE #113: BlockHeader doesn't roundtrip with serde binary encoding
Giving this error:
```
Generic("missing field `solution`")
```

I am working on a PR for this

RCasatta:
close #113 

The number of fields in the Params struct is wrong, this causes an error in roundtrip as shown in the included tests.

Strangely enough, this looks to happen only for binary encoding

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ elementsd = { version="0.3.0", features=["0_21_0","bitcoind_22_0"], optional = t
 rand = "0.6.5"
 serde_test = "1.0"
 serde_json = "1.0"
+serde_cbor = "0.8"  # older than latest version to support 1.36
 ryu = "<1.0.5"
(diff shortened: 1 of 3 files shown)


sanket1729 (approved):
ACK ad490d1da74fce6c1c111bdee8ddf6ce2763df9e. Verified that the round trip tests fail before the fix and pass successfully after the bug fix.

sanket1729 (approved):
ACK [0881ad4](https://github.com/ElementsProject/rust-elements/commit/0881ad49b322b56162aa57a83984f46950c03c66). Ran into the same issue while trying to use rust-secp-zkp last week. I think we will have to wait a couple of weeks for the rust-bitcoin release cycle to settle in before it is properly addressed.

PR #117, opened 2022-04-15
Add test vector for issuance with previous utxo with vout = 1

sanket1729:
Tests for vout os serialized correctly in the correct byte order. vout = 0 would result in the same ordering for LE and BE

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,9 +10,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


roconnor-blockstream:
Where did you get this test data from?

sanket1729:
This was created manually using elements rpc in regtest.

apoelstra (approved):
ACK cf2106a9015d2e422c270be50ba49aba87d9e052

PR #121, opened 2022-05-10
Add taproot psbt fields according to BIP371

sanket1729:
These work directly with elements 23.0(24.0?) after it has been released.
There are no changes to elements specific proprietary fields, only main
bitcoin psbt fields.

Most of this is just copy pasting code from rust-bitcoin and making it work :) . For reference: 
https://github.com/bitcoin/bips/blob/master/bip-0371.mediawiki. 

Unfortunately, this is the only supporting implementation. So, there are no test vectors to check against.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,9 @@ json-contract = [ "serde_json" ]
 "fuzztarget" = []
 
 [dependencies]
-bitcoin = "0.27"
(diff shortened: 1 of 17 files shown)


apoelstra (approved):
ACK 65f7838285308dc283918b62e13ccc270811f4b2

Did not confirm that the code maches the spec or rust-bitcoin, only that the code itself seemed sane. Realistically we will need to wait for another implementation to be able to meaningfully test/review the logic.

PR #111, opened 2021-12-22
Taproot Support

sanket1729:
Commits 86979bc518e781ed8c9bf19412652792c739fa5e and 16ba676ab8157bc823b08e6d5bdb5e1bbbdc5f26 don't compile. They are only separated for review aid. These commits have code that is copied from rust-bitcoin, and the subsequent commits make the elements specific changes which should be easy to review.

After this is ready for merge, I will squash the non-compiling commits to have a good history.

The last commit has integration tests for sighash/taproot tests for elements

PR #123, opened 2022-05-16
Add Support for Psbt Sighash type

sanket1729:
Also adds FromStr/Display for SchnorrSignatureHashType

- The first commit adds PsbtSighashType(needed for schnorr sigh psbt support)
- The second commit fixes a bug introduced in copy-pasting during the #111.
- The third commit does renaming to distinguish SchnorrSigHashType and EcdsaSigHashType
- The last and fourth commit fixes a small nit by adding a #[test] to untested code in serde feature.


src/pset/map/input.rs
```diff
@@ -258,6 +260,95 @@ pub struct Input {
     pub unknown: BTreeMap<raw::Key, Vec<u8>>,
 }
 
+/// A Signature hash type for the corresponding input. As of taproot upgrade, the signature hash
+/// type can be either [`SigHashType`] or [`SchnorrSigHashType`] but it is not possible to know
+/// directly which signature hash type the user is dealing with. Therefore, the user is responsible
+/// for converting to/from [`PsbtSighashType`] from/to the desired signature hash type they need.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
stevenroose, line 268: This derived will be a struct with an inner field that is a number, I doubt that is what we'd want. I think the other sighash types either use the `Display`/`FromStr` to serde-serialize as string, or just serialize as the number.

src/pset/map/input.rs
```diff
@@ -258,6 +260,95 @@ pub struct Input {
...
+pub struct PsbtSighashType {
+    pub (crate) inner: u32,
```
stevenroose, line 270: Specific reason why here you do a `pub (crate) inner: u32` and for the `NonStandardSighashType`, you just have a newtype with `pub u32`?

src/pset/map/input.rs
```diff
@@ -351,6 +351,30 @@ impl PsbtSighashType {
 
 impl Input{
 
+    /// Obtains the [`EcdsaSigHashType`] for this input if one is specified. If no sighash type is
+    /// specified, returns [`EcdsaSigHashType::All`].
+    ///
+    /// # Errors
+    ///
+    /// If the `sighash_type` field is set to a non-standard ECDSA sighash value.
+    pub fn ecdsa_hash_ty(&self) -> Option<EcdsaSigHashType> {
```
stevenroose, line 360: Just a nit, goes also for the method on `PsbtSighashType`, but it seems weird to try save 2 characters and shorten `type` to `ty`..

stevenroose (commented):
I guess mostly nits :)

src/pset/map/input.rs
sanket1729: No. Just an artifact copied from rust-bitcoin.

src/pset/map/input.rs
sanket1729: Would like to keep the names from rust-bitcoin here.

src/pset/map/input.rs
sanket1729: Indeed, I will add this in a followup commit to this PR

src/pset/map/input.rs
apoelstra: @stevenroose it's not to save characters, it's because `ty` is a keyword and therefore can't be used for variable names .. so we are trying to keep the naming consistent between variables and methods.

apoelstra (approved):
ACK 35110f6b5371319100913e6a6abed15f4637fe13

Though I agree with Steven's nit about the serde encoding. (But maybe this is copied from bitcoin and therefore ok?)

PR #124, opened 2022-05-18
Fix Block decoding bug

sanket1729:
SurjectionProofs were allocated on the stack for the entire array of characters even when they were empty. The fixes this by adding a Box around it.
This is not an issue for rangeproofs as there is already a Box in the ffi layer, but when that changes as we add more support to split range proof from a blob into individual components, we should make sure that we are boxing it here. 

While we are at it, this also fixes the issue for Psbt decoding while dealing with these cases.

The second commit can be swapped to see that test fails without the fix.


sanket1729:
Squashed the two commits

apoelstra (approved):
ACK d27565cde198654d6c7c24e99bd8e73adffccaf1

RCasatta:
Are we also going to Box the SurjectionProof upstream in the sys crate?

sanket1729:
> Are we also going to Box the SurjectionProof upstream in the sys crate?

My thinking currently is that we should not. The fact that we store SurjectionProofs and RangeProofs as optional is a rust-elements problem. Not a problem for rust-secp-zkp(-sys). On the contrary, I would expect the current implementation of Rangeproof(that uses Box<[uchar]>) to shift to an array once we split the blob into detailed splits. 

There are applications for these things beyond rust elements, and we should not the issues there affect secp-zkp-sys crate

apoelstra:
Agreed. I'd also like to avoid allocation in -sys if at all possible.

PR #126, opened 2022-06-01
Add TxIn::pegin_prevout method to return Bitcoin prevout

stevenroose:
This kind of makes the case "safer" for users.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -402,18 +402,14 @@ impl<'tx> PeginData<'tx> {
     /// Returns None if not a valid pegin witness.
     pub fn from_pegin_witness(
         pegin_witness: &'tx [Vec<u8>],
-        prevout: OutPoint,
+        prevout: bitcoin::OutPoint,
     ) -> Result<PeginData<'tx>, &'static str> {
         if pegin_witness.len() != 6 {
             return Err("size not 6");
         }
 
         Ok(PeginData {
-            // Cast of an elements::OutPoint to a bitcoin::OutPoint
-            outpoint: bitcoin::OutPoint {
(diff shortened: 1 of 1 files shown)


src/transaction.rs
```diff
@@ -295,11 +308,7 @@ impl TxIn {
         );
 
         Some(PeginData {
-            // Cast of an elements::OutPoint to a bitcoin::OutPoint
-            outpoint: bitcoin::OutPoint {
-                txid: bitcoin::Txid::from(self.previous_output.txid.as_hash()),
-                vout: self.previous_output.vout,
-            },
+            outpoint: self.pegin_prevout().expect("this is a pegin input"),
```
sanket1729, line 311: nit: You can also `?` here(since it is can already fail) to avoid future readers of the code thinking about this.

sanket1729 (approved):
ACK 1195a43bdaaf68eef4482d0dbec860aaf0e5c4b7. Really a nit, feel free to ignore.

src/transaction.rs
apoelstra: Could also change the top `if self.is_pegin` check to an `if let` on this data.

stevenroose: addressed these nits

stevenroose: ah damn had this locally fixed but didn't push that, my bad

sanket1729 (approved):
ACK e52a71fd0acf764883fe0fd31c11e43731523fda

apoelstra (approved):
ACK 2085ad4eeddc92f40bd20847c93fc6362b5c964f

PR #128, opened 2022-06-07
Deprecate get_ prefix for Block and Transaction methods

shesek:
To match the changes introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/861.

This allows using unified code with the same methods for both `elements::{Transaction,Block}` and `bitcoin::{Transaction,Block}` (behind different features), without triggering rust-bitcoin's deprecation warnings for the `get_`-prefixed methods.

Note that there are other instances of `get_`-prefixed methods in the rust-elements codebase. I only updated the ones needed for compatibility with the changes made in rust-bitcoin.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -372,7 +372,13 @@ impl Block {
     }
 
     /// Get the size of the block
+    #[deprecated(since = "0.19.1", note = "Please use `Block::size` instead.")]
     pub fn get_size(&self) -> usize {
+        self.size()
+    }
+
+    /// Get the size of the block
+    pub fn size(&self) -> usize {
         // The size of the header + the size of the varint with the tx count + the txs themselves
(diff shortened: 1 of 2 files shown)


src/block.rs
```diff
@@ -372,17 +372,29 @@ impl Block {
     }
 
     /// Get the size of the block
+    #[deprecated(since = "0.19.1", note = "Please use `Block::size` instead.")]
```
shesek, line 375: This assumes that the next release is going to be `0.19.1`, which might not be the case.

apoelstra (approved):
ACK 2a35bad194979ae511949322b1c59ff2e37dd5e6

apoelstra:
I'm happy to do a quick point release if you want. There's nothing else in the queue.

shesek:
No urgency for me, I can keep using the compatibility trait until a new release is out. Thanks for merging!

PR #129, opened 2022-06-09
Add some default impls

sanket1729:
These are the same values from rust-bitcoin. Makes it easy to create
these structures downstream. Right now, we need to specify all feilds


delta1:
ack 34cf67c0f831e5387208a5fef93e13c80cef8937

PR #130, opened 2022-06-09
encode: Revert io::BufRead requirement on Encodable trait

stevenroose:
We really never should have required `BufRead` just to make the commitment parsing a tad more convenient.

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -181,8 +181,7 @@ impl Decodable for Value {
                 let mut comm = [0u8; 33];
                 comm[0] = p;
                 d.read_exact(&mut comm[1..])?;
-                let commitment = Decodable::consensus_decode(&comm[..])?;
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 203542c902f4975c237ec2985c502469872dbde9

stevenroose:
Btw, I think this change can be considered non-breaking. I don't remove the now-unused trait impls and BufRead is a struct supertrait of Read.

apoelstra (approved):
ACK c1a7641ff38d897390c9a7e054621e5ef7de598a

sanket1729 (approved):
ACK c1a7641ff38d897390c9a7e054621e5ef7de598a. `Commitment` and `Generator` are Copy. There is no allocation in them. I would like to keep the decodable Impls because they can come in handy in downstream repos.

PR #139, opened 2022-07-13
Avoid panic malformed pegin

stevenroose:
Noticed while reading that a malformed pegin witness could actually make our code panic.


apoelstra:
Good catch!

PR #143, opened 2022-08-03
Blinding API cleanups

sanket1729:
- This required the user to explicitly maintain whether the current TxIn
has issuance or not whereas it can directly be calculated from
assetIssuance field

The blinding APIs are separated into smaller chunks. This allows
    - Blinding only the values or the assets but not both
    - Not relying on rng to set asset/value blinding factors
    - Better input parameters

Note I had to edit the test vectors the order in which they were sampled from rng had changed.


apoelstra:
Should I review this? I have been holding off because of the sea of red in CI.

sanket1729:
The downstream options project is stable enough. I can rewive this PR today.

sanket1729:
Cherry-picked the MSRV commit here. This might not be an easy review, but I would really like this to go into the next release.

apoelstra:
Will need rebase after #147 (and then I think we'll no longer need the MSRV bump, though we will need a CHANGELOG update). But I'll review this as-is.

src/pset/map/input.rs
```diff
@@ -420,9 +421,33 @@ impl Input{
...
+            let prevout = OutPoint {
+                txid: self.previous_txid,
+                vout: self.previous_output_index,
+            };
+            let contract_hash =
+                ContractHash::from_inner(self.issuance_asset_entropy.unwrap_or_default());
+            AssetId::generate_asset_entropy(prevout, contract_hash)
+        } else {
+            // re-issuance
+            sha256::Midstate::from_inner(self.issuance_asset_entropy.unwrap_or([0u8; 32]))
```
apoelstra, line 439: In 8b41c1327d8af43f2196ce97d230c5a0719d74c3:

Should use `unwrap_or_default` here like above (or change both to `unwrap_or([8; 32])`; having two ways to write the same thing makes it look like these  do different things, which is confusing.

src/blind.rs
```diff
@@ -239,37 +239,161 @@ impl TxOutSecrets {
...
+/// while inputs from user's wallet should be [`SurjectionInput::Known`]
+///
+/// Explicit assets can be provided as [`SurjectionInput::Unknown`]. There is no
+/// need to construct a `Known` variant with secrets
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum SurjectionInput{
+    /// Unknown inputs for whom we don't know the secrets(asset tags/blinding factors)
+    Unknown(Asset),
+    /// Known inputs for whom we know corresponding [`TxOutSecrets`]
+    Known(TxOutSecrets),
```
apoelstra, line 268: In 8b41c1327d8af43f2196ce97d230c5a0719d74c3:

I think this should be
```
Known {
    asset: AssetId,
    asset_bf: AssetBlindingFactor,
}
```
These fields are both `Copy` and small-ish and easy to copy out of `TxOutSecrets`. This would make this type tighter -- right now you use `TxOutSecrets` and sometimes just put dummy data into them which is potentially confusing.

src/blind.rs
```diff
@@ -15,12 +15,12 @@
 //! # Transactions Blinding
 //!
 
-use std::{self, fmt};
+use std::{self, fmt, collections::BTreeMap};
```
apoelstra, line 18: In 8b41c1327d8af43f2196ce97d230c5a0719d74c3:

I think you should use `HashMap` everywhere instead of `BTreeMap`. It doesn't really make a difference at all but you never iterate over this map directly (you always iterate over the Vec of tx inputs) and `HashMap` is a little bit easier to use and potentially more efficient.

apoelstra (approved):
ACK 49d1ec2719cc4eef76359676fb223a15f8a7f95b

apoelstra:
All my comments are nits which could be addressed with additional commits or even in another PR.

Also, can you add a commit which runs `cargo fmt`?

sanket1729:
> Also, can you add a commit which runs cargo fmt?

@apoelstra, do you mean on the entire codebase

apoelstra:
@sanket1729 let's say, just on blind.rs and psbt/mod.rs which are otherwise heavily modified in this PR (and there are a couple format nits I didn't want to write out in those files).

sanket1729:
@apoelstra, any interest in reviving https://github.com/ElementsProject/rust-elements/pull/80? We have enforced `rustfmt` across rust-bitcoin org in many repos.

sanket1729:
Addressed the comments. Will rebase we have #147 in

sanket1729:
Also added a test to test tx verification with partially blinded transactions

src/pset/mod.rs
```diff
@@ -352,14 +347,55 @@ impl PartiallySignedTransaction {
...
-            .filter(|o| o.is_some())
-            .map(|o| o.unwrap())
-            .map(|o| (o.value, o.asset_bf, o.value_bf))
+            .map(|(_i, sec)| (sec.value, sec.asset_bf, sec.value_bf))
             .collect::<Vec<_>>();
 
         Ok((inp_secrets, blind_out_indices))
     }
 
+    /// Obtains the surjection inputs for this pset. This servers as the domain
```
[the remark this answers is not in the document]
apoelstra: Still present in 04ad35f

src/pset/map/input.rs
apoelstra: Still present in 04ad35f

apoelstra (approved):
ACK 073e720180bc139d4d1c89b4d8a6dee122a8a13e

src/pset/map/input.rs
sanket1729: Addressed in fa753b4

src/pset/mod.rs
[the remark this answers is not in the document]
sanket1729: Addressed in fa753b4

apoelstra (approved):
ACK 582325ef3dae38a3f51ebe237697d181980b0ec4

PR #144, opened 2022-08-24
Add utility methods to create pegin witnesses

stevenroose:
This is a breaking change only in the field of the asset. I thought it made more sense like that, but if we prefer to not make breaking changes, the utility methods could also be added without the field change.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -184,6 +184,50 @@ pub struct PeginData<'tx> {
     pub referenced_block: bitcoin::BlockHash,
 }
 
+impl<'tx> PeginData<'tx> {
+    /// Construct the pegin data from a pegin witness.
+    /// Returns None if not a valid pegin witness.
+    pub fn from_pegin_witness(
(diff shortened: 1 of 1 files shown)


stevenroose:
Even though this seems to pass unit tests, I'm personally experiencing some issues with the assetid not roundtripping correctly. I'll try remember posting an update when I have that figured out.

apoelstra:
Ok, will hold off on reviewing until then. If it turns out to a bug please add a unit test :)

stevenroose:
Ah damn I had this comment written but didn't submit yet..

Ok, nevermind; the issue was on my side because I copied over the impls but I ofc still had the old field type so it was serializing the confidential asset with prefix and trying to deserialize just the plain asset id.

It should be good to go now

apoelstra (approved):
ACK b5cfed432218251c7b9720fd482e2f004a446b7b

sanket1729 (approved):
utACK b5cfed432218251c7b9720fd482e2f004a446b7b

PR #145, opened 2022-08-30
fix fuzzing

apoelstra:
Fuzzing does not work locally for me because a dep of honggfuzz broke compilation. We can disable it by disabling optional features.


sanket1729 (approved):
utACK ea1676e96035b04cffe0349d62c6d4d415694d7d

PR #146, opened 2022-09-07
Export Blind*Proofs traits

LeoComandini:
So that the traits methods can be used downstream.


apoelstra (approved):
ACK ebc6b8e96d48ca8bc8721dcb87317162c84253f1

PR #152, opened 2022-10-18
Introduce Sequence and LockTime from rust-bitcoin

apoelstra:
The PSET stuff I'm a little uncertain about -- PSBT2 brings new locktime logic in that doesn't exist in PSBT0. My new logic is essentially the same, except that while previously we would allow the user to specify invalid locktimes (since we just parsed a u32 with no further checking) now we don't.

I think, when we do this upstream in rust-bitcoin, we should probably introduce some helper methods into LockTime, because currently there are conversions from Time/Height into LockTime that feel misplaced.

Also, it is really unclear what the difference between LockTime and PackedLockTime is supposed to be, so I did my best to copy rust-bitcoin.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,13 +44,16 @@ mod block;
 pub mod confidential;
 pub mod dynafed;
 pub mod encode;
+mod error;
 mod fast_merkle_root;
 pub mod hash_types;
+pub mod locktime;
 pub mod issuance;
 pub mod opcodes;
 pub mod script;
 mod transaction;
 mod blind;
+mod parse;
 pub mod slip77;
 pub mod sighash;
 pub mod pset;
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -62,6 +62,9 @@ mod serde_utils;
 mod endian;
 // re-export bitcoin deps which we re-use
 pub use bitcoin::{bech32, hashes};
+// re-export bitcoin locktime types, which have identical semantics as in Elements,
+//  and differ only by what kind of transaction they appear in.
+pub use bitcoin::{blockdata::locktime, LockTime, PackedLockTime, Sequence};
```
sanket1729, line 67: Would need to double-check that having different block rates(1min vs 10 mins) does not mess up the APIs here. I would really like to have separate `Sequence` and `LockTime` types here because we can possibly change semantics to mean different things. Initially, we used bitcoin::Script here but found it better to use elements-specific types in the long run.

apoelstra: Ok, I'll make them distinct types. I agree that Elements blockheights are semantically different from Bitcoin blockheaights.

Would it be OK if I just did newtypes around the rust-bitcoin types, and passed through every method?

sanket1729: I think we should copy the lock time and sequence module from rust-bitcoin instead of wrapping new types. Wrapper types might make the API annoying to use. The modules are self-contained and hopefully should be easy to copy-paste. 

We did the same thing for `Script` and even decided it was for keys. But we lose the benefit of gaining bug fixes and updates from upstream for free.

sanket1729: Overall, I think defining all primitives in `rust-elements` will make things easy in the future as 
1) I don't expect the upstream locktime/sequence modules to change often
2) When we have the CSV  for relative timelocks in blocks fixed in elements/liquid, we can change things here.

apoelstra:
Added a couple commits which copy more code in from rust-bitcoin.

sanket1729 (approved):
ACK ee76903b753c8a103c81aa0208246deb42fe529f

PR #154, opened 2022-10-19
Add PeginData::mainchain_tx inspector

stevenroose:
Just a nice utility for the mainchain tx.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+
+# 0.21.1 - 2022-10-21
+
+- Add `PeginData::parse_tx`
+- Add `PeginData::parse_merkle_proof`
+
 # 0.21.0 - 2022-10-19
 
 - Copy `Sequence` and `LockTime` structures from rust-bitcoin 0.29.1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "elements"
-version = "0.21.0"
+version = "0.21.1"
 authors = ["Andrew Poelstra <apoelstra@blockstream.com>"]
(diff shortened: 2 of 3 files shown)


apoelstra:
If you add a commit to do a point release we could get this out in 0.21.1.

sanket1729 (approved):
utACK ad26068e7a5ec7da19d485c30bb6f5efe60bcab3

stevenroose:
Ok, I'll also add the same getter for the merkle proof because I need it too :)

stevenroose:
I changed the name, so sanket's ack is void, sorry. Also added extra commit and point release.

I based the conclusion that the merkle_proof is a merkleblock based on the fact that we fill with with `gettxoutproof` output and that one has
```cpp
    CDataStream ssMB(SER_NETWORK, PROTOCOL_VERSION | SERIALIZE_TRANSACTION_NO_WITNESS);
    CMerkleBlock mb(block, setTxids);
    ssMB << mb;
    std::string strHex = HexStr(ssMB);
    return strHex;
```

sanket1729 (approved):
utACK 72afb4740e26c9cd658a0d544e7579f329f2b808. Looks good, but would be good to have some fixed test vector here.

apoelstra (approved):
ACK 72afb4740e26c9cd658a0d544e7579f329f2b808

apoelstra:
@stevenroose yep, it's a merkleblock -- as you noticed, we ask users to generate this using `gettxoutproof` (well, I think we do this automatically usually).

I would also like a fixed test vector but I don't think we need to hold up the PR or release for it. I'd be okay doing a 0.20.2 "bugfix" release on the off chance that this is somehow broken.

apoelstra:
Tagged & published

PR #159, opened 2023-04-10
Tap tweak

uncomputable:
Updates `TapTweak` to [latest rust-bitcoin](https://docs.rs/bitcoin/latest/bitcoin/key/trait.TapTweak.html) and adds `TweakedKeyPair`.

### DIFF
--- a/src/schnorr.rs
+++ b/src/schnorr.rs
@@ -19,12 +19,10 @@
 
 use std::fmt;
 
-use secp256k1_zkp::Scalar;
-pub use secp256k1_zkp::{XOnlyPublicKey, KeyPair};
-use secp256k1_zkp::{self, Secp256k1, Verification, constants::SCHNORR_SIGNATURE_SIZE};
-use crate::hashes::{Hash, HashEngine};
 use crate::taproot::{TapBranchHash, TapTweakHash};
 use crate::SchnorrSigHashType;
+use secp256k1_zkp::{self, constants::SCHNORR_SIGNATURE_SIZE, Secp256k1, Verification};
+pub use secp256k1_zkp::{KeyPair, XOnlyPublicKey};
 
 /// Untweaked Schnorr public key
 pub type UntweakedPublicKey = XOnlyPublicKey;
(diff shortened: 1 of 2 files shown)


uncomputable:
We could also update the methods of `TweakedPublicKey` if we want.

uncomputable:
The `syn-2.0.13` dependency seems to use the 2021 edition, which is not understood by rust 1.41.1.

src/schnorr.rs
```diff
@@ -33,38 +31,63 @@ pub type UntweakedPublicKey = XOnlyPublicKey;
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub struct TweakedPublicKey(XOnlyPublicKey);
 
-/// A trait for tweaking Schnorr public keys
+/// A trait for tweaking Schnorr key types (x-only public keys and key pairs).
 pub trait TapTweak {
```
apoelstra, line 42: In 063a8f5fffe90e8bd729c10b254825dbb4420076:

Why do we even have this trait? Why not use the one from Bitcoin?

apoelstra: Same with this entire module, really.

uncomputable:
The difference between this trait and the one from rust-bitcoin is that tweaking is done with `elements::TapBranchHash`. This little detail is really annoying when working with `elements::taproot::TaprootSpendInfo`.

apoelstra:
Gotcha, ok.

Lemme spend a bit more time thinking if there's a better way to keep these crates in sync. But maybe not, and this PR should go in.

apoelstra (approved):
ACK 7b3f081dc2a20769ca9b467b0472c0cda49ef0c7

apoelstra:
Casting hashtypes would not be sufficient to eliminate the code duplication. In general, we need to do an overhaul of this library to bring it closer to rust-bitcoin, but Sanket and I decided to wait for rust-bitcoin 1.0 (or at least, for it to quiet down a lot) before trying this.

PR #160, opened 2023-04-10
Make TxOut generic in Prevouts

uncomputable:
Useful when prevouts are saved in a vector.

### DIFF
--- a/src/sighash.rs
+++ b/src/sighash.rs
@@ -19,6 +19,7 @@
 //! signatures, which are placed in the scriptSig.
 //!
 
+use std::borrow::Borrow;
 use crate::encode::{self, Encodable};
 use crate::hash_types::SigHash;
 use crate::hashes::{sha256d, Hash, sha256};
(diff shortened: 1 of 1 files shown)


sanket1729:
I guess MSRV broke :(

uncomputable:
`cargo +1.41.1 test --verbose --features=serde` seems to be broken. Since this is the only feature currently tested, I suggest bumping the Rust version to something like 1.58 for the single serde test. `integration` seems to be broken, too. I can make a PR.

RCasatta:
Didn't look the details but recently serde upgraded syn which broke MSRV in another repo of mine.
If it's the case I think adding `cargo update -p serde --precise 1.0.152` in the CI with toolchain `1.41.1` is more appropriate

uncomputable:
@RCasatta Thanks for pointing this out. @apoelstra did something along these lines in #162. CI seems to be working now.

apoelstra:
dc6f5202ed276511cee002cd5942c28cd58a9f32 looks good. Verified that it is essentially the same as 10fedfb3b44edef1aa69c16452fd27bb8cf65309 in rust-bitcoin.

apoelstra (approved):
ACK dc6f5202ed276511cee002cd5942c28cd58a9f32

Issue #110, opened 2021-12-10
Reintroduce psbt_rtt check once upstream elements fixes decodepsbt for PSBT_WITNESS_UTXO 


sanket1729:
upstream issue: https://github.com/ElementsProject/elements/issues/1080
Current discussion: ElementsProject/rust-elements#108 (comment)

PR #162, opened 2023-04-12
separate out elementsd tests into separate crate; allow overriding the binaries

apoelstra:
I need this to test this crate within Nix.


sanket1729 (approved):
ACK 2d14ff7c9b48d8a2424078df2bd77e36e601a36f. We can debug `tx-pegin`  test later

sanket1729:
Looks like the test was ignored previously too?

sanket1729:
Right, looking at it. It is an elementsd bug.
```
#[ignore] // TODO this fails because elements decodepsbt is not printing TxOut::asset (PSET_IN_WITNESS_UTXO)
```

sanket1729:
We can later update that test to test something else.

sanket1729:
Right I even had #110

PR #164, opened 2023-05-11
Remove std feature gates

uncomputable:
There is no `std` feature in rust-elements. This is a copy-and-paste error from rust-bitcoin.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,14 +7,10 @@ pub use crate::parse::ParseIntError;
 macro_rules! impl_std_error {
     // No source available
     ($type:ty) => {
-        #[cfg(feature = "std")]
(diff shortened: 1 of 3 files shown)


uncomputable:
> error: There are multiple `syn` packages in your project, and the specification `syn` is ambiguous.

I feel this is not from my PR. CI broke again?

apoelstra (approved):
ACK b14f1a31c5a9c7244b965e7c58e139798a0c9c64

apoelstra:
@uncomputable yes, CI is broken because of `syn` (which I thought we'd fixed many weeks ago..) and because we need doc comments on `pub extern crate` now (this one is new).

PR #166, opened 2023-06-08
Update crate to bitcoin 0.30

apoelstra:
Apologies for taking so long. Every time I got started on this over the last 6 months I'd run into problems with hex encoding immediately and then stall out.

I tried to break this into reasonably-digestible chunks. Each commit should compile and pass all tests independently. All the scaffolding should be gone by the end of the PR.

I also updated the CI since the old version was 404ing, and while I was at it I added tests for documentation and clippy, so I needed a few more commits to fix that.


apoelstra:
Oops, I need to push a new version of secp-zkp first.

apoelstra:
Ok, all the CI jobs that are gonna run have passed :) we are good to go now.

contrib/test.sh
```diff
@@ -2,35 +2,51 @@
...
-then
-    alias cargo="cargo +$TOOLCHAIN"
-fi
-
 # Pin dependencies as required if we are using MSRV toolchain.
-if cargo --version | grep "1\.41"; then
+if cargo --version | grep "1\.48"; then
     # 1.0.157 uses syn 2.0 which requires edition 2018
     cargo update -p serde --precise 1.0.156
     # 1.0.108 uses `matches!` macro so does not work with Rust 1.41.1, bad `syn` no biscuit.
```
RCasatta, line 9: this comment about 1.41.1 is not relevant anymore

contrib/test.sh
```diff
@@ -2,35 +2,51 @@
...
 
-# Use toolchain if explicitly specified
-if [ -n "$TOOLCHAIN" ]
-then
-    alias cargo="cargo +$TOOLCHAIN"
-fi
-
 # Pin dependencies as required if we are using MSRV toolchain.
-if cargo --version | grep "1\.41"; then
+if cargo --version | grep "1\.48"; then
```
RCasatta, line 6: I have suspect toolchain 1.48 is not really running... Locally with 1.449 I have errors on `which` and I think it needs at least:

```
cargo update -p which --precise 4.3.0
```

contrib/test.sh
apoelstra: Good catch, now we only need the serde pin.

contrib/test.sh
apoelstra: Can you post your exact invocation? `cargo +1.48.0 test --all-features --all-targets` runs clean for me with only the serde pin.

apoelstra: Ok, if I do `cargo +stable update` then it pulls in a version of `once_cell` thatt 1.48.0 doesn't recognize, and which gives me an error message blaming `which` for some reason. But if I do `cargo +1.48.0 update` then this doesn't happen.

So I think we're fine.

apoelstra:
Removed unnecessary `syn` pin and changed CI to use @RCasatta's `on:` recipe. No other changes.

sanket1729 (approved):
ACK 74948ea9fdd7ba718826d3a2f1a02dd9e3ad35d1. Thanks for doing this in steps with smaller commits. Made this PR really easy to review.

sanket1729:
I will update branch protection rules after this is merged.

PR #171, opened 2023-08-02
Create explicit empty and null values for some types

stevenroose:
IMO these are different than f.e. the Default impls on TxOut etc, because these types actually in many actual production cases want to be set to the empty or none value. As opposed to Default on other types that only exist to aid in test scenarios or so to avoid having to set the fields.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -34,7 +34,7 @@ use secp256k1_zkp::{
 };
 
 /// Description of an asset issuance in a transaction input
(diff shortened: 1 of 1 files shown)


delta1 (approved):
ack 349de415d2edbe945d9e651e7c5e8a56d50dd151

apoelstra (approved):
ACK 349de415d2edbe945d9e651e7c5e8a56d50dd151

PR #172, opened 2023-08-02
Change error type of Encodable to io::Error like bitcoin does

stevenroose:


stevenroose:
Really? Clippy in CI??

I disagree with clippy's suggestion in this case, though. I think 
```rust
Ok(method_with_same_error_type()?)
```
is more expressive than leaving out the Ok and just forwarding. It also is more robust as it will probably not break if you change the signature of either the parent or child method. I think this is a dumb clippy suggestion, tbh.

sanket1729:
Feel free to ignore specific rules #[allow(clippy::rule)].

sanket1729:
Adding it in CI has served us well to spot some extra allocations, matching against wildcard etc. Overall, I am of the opinion that we should keep it in CI and ignore specific rules if we think are bothering us.

apoelstra:
I agree with clippy's suggestion but I really don't care, it's just a style rule. Feel free to ignore it to get CI to pass

apoelstra:
Closing. This is out of date now.

PR #173, opened 2023-08-02
Fix examples

sanket1729:
Broken in 6d67a6c. AssetId is actually parsed reverse.


delta1 (approved):
ack 34baadab4446ccde26419a043b3763cf1ddfc9e7

apoelstra (approved):
ACK 34baadab4446ccde26419a043b3763cf1ddfc9e7

Issue #105, opened 2021-11-24
Incompatible Value binary serialization format


shesek:
It appears that the [bincode value fix](#98) resulted in a different serialization format compared to before the [regression](#96) (v0.16).

With current master (fe3e9469618a479de02f8ececf92f9d456fc4c38), the following code:

```rust
let commitment = PedersenCommitment::from_slice(&Vec::from_hex("09b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4").unwrap()).unwrap();
let value = Value::Confidential(commitment);

println!("BE bytes: {}", bincode::serialize(&value).unwrap().to_hex());
println!("LE bytes: {}", bincode::DefaultOptions::default().with_little_endian()
    .serialize(&value).unwrap().to_hex());
```

Results in the following serialization:

```
BE bytes: 020000000000000002210000000000000009b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4
LE bytes: 02022109b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4
```

While with v0.16, the following code:

```rust
let value = Value::from_commitment(&Vec::from_hex("09b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4").unwrap()).unwrap();

println!("BE bytes: {}", bincode::serialize(&value).unwrap().to_hex());
println!("LE bytes: {}", bincode::DefaultOptions::default().with_little_endian()
    .serialize(&value).unwrap().to_hex());
```

Results in the following serialization:

```
BE bytes: 020000000000000009b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4
LE bytes: 0209b88680778d3d5530785ff12bb62e6ac98c7be5a8b9b0ea392e166bf61ea77ee4
```

stevenroose:
(The links in the first line don't work.)

stevenroose:
Seems like https://github.com/ElementsProject/rust-elements/pull/77 introduced this. We definitely could use more tests where all serializations are tested for regressions.

shesek:
> (The links in the first line don't work.)

Oops, sorry, fixed them now.

shesek:
[...]

PR #174, opened 2023-08-21
confidential: Don't reverse explicit values for serde

stevenroose:
So this is a breaking change. It might affect people actually using the serde serialization of `confidential::Value::Explicit` right now. Which is really unfortunate.

But IMO this is a big. We for some miraculous reason (probably a mistake on some programmer's end) reverse the bytes when consensus-encoding explicit values. That's fine. But somehow we also ended up doing this for serde, which makes no sense. 

I'm hitting this when trying to use this type in a Web setting through WASM and a small number of satoshis will (when reversed) cause a JS overflow (which only supports 53 bit integers).

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -202,7 +202,7 @@ impl Serialize for Value {
             Value::Null => seq.serialize_element(&0u8)?,
             Value::Explicit(n) => {
                 seq.serialize_element(&1u8)?;
-                seq.serialize_element(&u64::swap_bytes(n))?;
+                seq.serialize_element(&n)?;
             }
             Value::Confidential(commitment) => {
                 seq.serialize_element(&2u8)?;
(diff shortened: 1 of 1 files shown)


stevenroose:
Maybe I should use the opportunity to change the serde for all confidential types to be a bit more JSON-friendly and use the lenght/type of the field to know whether it's confidential or not... Opinions welcome.

apoelstra:
Why doesn't it make sense for the serde encoding to match the on-wire encoding for explicit amounts, when it matches the encoding for confidential amounts?

sanket1729:
I agree that this is more of a bug, but I think that we should just live with it. This change will break a lot of things in many subtle ways. I think some implementations for liquid of electrum server also use serde format as a persistent store. 

Can you work around this for your setting?

sanket1729:
See https://github.com/ElementsProject/rust-elements/pull/96 and https://github.com/ElementsProject/rust-elements/issues/105 where we broke serde format across versions.

apoelstra:
If it's a bug, it seems like it's a consensus bug in Elements. We can fix it during the bulletproof hardfork maybe.

apoelstra:
Closing because this is a breaking serialization change and we don't have consensus on moving forward.

PR #185, opened 2024-01-03
Exposed RangeProofMessage publically

i5hi:
The `RangeProofMessage` stuct needs to be publically exposed for usage of [confidential::Value::blind](https://docs.rs/elements/latest/elements/confidential/enum.Value.html#method.blind)


apoelstra (approved):
ACK 386e3b8ecf84b1162706c73e3db980fda30903ff

PR #186, opened 2024-01-03
Updated doc for impl Value blind method - returns blinded value*

i5hi:
Fixes a typo in Value::blind doc


apoelstra (approved):
ACK 2441f171e0188c846c0cab70b0e9c9ab4effa8d2

PR #188, opened 2024-01-03
Update rust-bitcoin to 0.31.0, and associated dependencies

apoelstra:
Let me know if I should split this into two PRs; the blech32 changes should arguably be done separately.


apoelstra:
Oops, CI failure is that rust-secp-zkp is broken with `cfg=fuzzing`. I need to do another point release.

apoelstra:
Blocked on https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/76 which will fix CI here.

src/taproot.rs
```diff
@@ -23,55 +23,36 @@ use std::collections::{BTreeMap, BTreeSet, BinaryHeap};
...
-);
-sha256t_hash_newtype!(TapTweakHash, TapTweakTag, MIDSTATE_TAPTWEAK, 64,
-    doc="Taproot-tagged hash for elements public key tweaks", forward
-);
-sha256t_hash_newtype!(TapSighashHash, TapSighashTag, MIDSTATE_TAPSIGHASH, 64,
-    doc="Taproot-tagged hash for the elements taproot signature hash", forward
-);
+sha256t_hash_newtype! {
+    pub struct TapLeafTag = hash_str("TapLeaf/elements");
+    /// Taproot-tagged hash for elements tapscript Merkle tree leafs".
```
RCasatta, line 29: spurious end with a '"' here and in other doc comments of the hashes

src/taproot.rs
```diff
@@ -23,55 +23,36 @@ use std::collections::{BTreeMap, BTreeSet, BinaryHeap};
...
+    #[hash_newtype(forward)]
+    pub struct TapLeafHash(_);
+
+    pub struct TapBranchTag = hash_str("TapBranch/elements");
+    /// Tagged hash used in taproot trees; see BIP-340 for tagging rules", forwar
+    #[hash_newtype(forward)]
+    pub struct TapNodeHash(_);
```
RCasatta, line 36: Should be `TapBranchHash` ?

RCasatta:
In a downstream test, I've got an error on address generation which seems related only to the checksum.

```
thread 'tests::test_ks_flow' panicked at ks_bindings/src/lib.rs:312:9:          
assertion `left == right` failed                                                
  left: "tlq1qq2xvpcvfup5j8zscjq05u2wxxjcyewk7979f3mmz5l7uw5pqmx6xf5xy50hsn6vhkm5euwt72x878eq6zxx2z58hd7zrsg9qn"
 right: "tlg1qq2xvpcvfup5j8zscjq05u2wxxjcyewk7979f3mmz5l7uw5pqmx6xf5xy50hsn6vhkm5euwt72x878eq6zxx2z03a3gsu4psmv"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace 
```

I don't follow bech32 logic deeply, but I think the error is about the constant `0x455972a3350f7a1` which in this MR is used only to check (read-only) while on master is used also in creating the checksum https://github.com/ElementsProject/rust-elements/blob/613204f38eeb8f4fcc767042197fd763965f057a/src/blech32.rs#L195

src/taproot.rs
apoelstra: Upstream we renamed this to `TopNodeHash` in https://github.com/rust-bitcoin/rust-bitcoin/pull/1479 in order to simplify the single-leaf Taproot case.

apoelstra:
Thanks! I will investigate the checksum bug next week. I'm not sure what you mean by "used only to check". You can see in the `fmt::Display` impl of my new code that I use `Blech32`/`Blech32m` for encoding as well.

My guess is that I messed up the logic for choosing between bech32 and blech32.

RCasatta:
> I'm not sure what you mean by "used only to check"

I was meaning that the constant is not used to write another value, like it's done in master in the line `let plm: u64 = polymod(&values) ^ variant.constant(); ` but I realiazed this could happen in the bech32 crate which I didn't check. So disregard the comment.

jamesdorfman:
ACK 009205c5c3f7b4a91d28d2715f0edb62f1b8fb30. I don't fully understand the bech32 commit -- someone  more knowledgable than me should review that. But, the rust-bitcoin changes LGTM.

apoelstra:
Ok, should be good to go. Fixed the tlq/tlg thing, and CI should be fixed because we updated secp256k1-zkp to expose some trait impls even in fuzzing mode.

apoelstra:
Oh, derp, I need to release a new version of rust-secp-zkp, not just rust-secp-zkp-sys.

apoelstra:
cc https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/78

RCasatta:
>  The checksum thing was a red herring;

Doh! Sorry for being the first to fall into the trap and accusing the poor checksum to be at fault!

Going to check again downstream and report

RCasatta:
No more errors found downstream, LGTM

@apoelstra did you forget this? https://github.com/ElementsProject/rust-elements/pull/188#discussion_r1441771204 doccomments of that line and the following needs some trim

apart from that

ACK 7fe86e544de8e3ce585aaa1337fde0cc8af0039c

apoelstra:
Oops, I did forget. Fixed.

delta1 (approved):
utACK 092ae6163a989b69a15c3992afddbcdc8ad8cdce

RCasatta (approved):
ACK 092ae6163a989b69a15c3992afddbcdc8ad8cdce

PR #199, opened 2024-04-20
pset: input: add explicit amount and asset, and their proofs

LeoComandini:
They were added in https://github.com/ElementsProject/elements/pull/1157


LeoComandini:
`PSBT_ELEMENTS_IN_BLINDED_ISSUANCE` changes the behavior for `PartiallySignedTransaction::blind_*` if the pset has an issuance, so I will add that in a separate MR with proper tests

apoelstra (approved):
ACK eb85a23bb7afe69a952e57ae0464716334c2eff8

PR #200, opened 2024-04-22
pset: input: add blinded issuance flag

LeoComandini:
Added in https://github.com/ElementsProject/elements/pull/1157


apoelstra (approved):
ACK 908cb5eb53777964fe54a0d54011966df9c897ed this looks good for now. Would be good to actually implement the blinding of course

PR #201, opened 2024-04-23
pset: add optional asset blinding factor to input and output

LeoComandini:
Change corresponding to https://github.com/ElementsProject/elements/pull/1329


apoelstra (approved):
ACK 17d354eaf537d5ee79f27790568ce5ba037e6012

PR #203, opened 2024-06-18
transaction: range-check pegin data when parsing

apoelstra:
It is possible to cause a panic by calling `TxIn::pegin_data` on a malformed pegin, which has 6 entries and the last one (which should be a Merkle proof of block inclusion) is too small to even have a blockhash in it.

On the actual chain this should not be possible because invalid pegin data would be rejected.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -436,6 +436,9 @@ impl<'tx> PeginData<'tx> {
         if pegin_witness.len() != 6 {
             return Err("size not 6");
         }
+        if pegin_witness[5].len() < 80 {
+            return Err("merkle proof too short");
+        }
 
         Ok(PeginData {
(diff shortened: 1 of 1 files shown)


apoelstra:
CI failure is an unrelated thing; I have another branch I'm working on to fix clippy and update things but I'm seeing a weird bug with clippy.

delta1 (approved):
ACK f86e219086e25288e4508b02996c8058e3b36496

apoelstra:
Rebased. Ready for re-review.

delta1:
reACK [639c964](https://github.com/ElementsProject/rust-elements/pull/203/commits/639c964f91894bf5f22d7a3fe9ac504ab31a3c7d)

PR #206, opened 2024-07-02
pset: elip100: add and get token metadata

LeoComandini:
Following changes in https://github.com/ElementsProject/ELIPs/pull/17

### DIFF
--- a/src/pset/elip100.rs
+++ b/src/pset/elip100.rs
@@ -21,6 +21,9 @@ use crate::{
 /// keytype as defined in ELIP0100
 pub const PSBT_ELEMENTS_HWW_GLOBAL_ASSET_METADATA: u8 = 0x00u8;
 
+/// Token keytype as defined in ELIP0100
+pub const PSBT_ELEMENTS_HWW_GLOBAL_REISSUANCE_TOKEN: u8 = 0x01u8;
+
 /// Prefix for PSET hardware wallet extension as defined in ELIP0100
 pub const PSET_HWW_PREFIX: &[u8] = b"pset_hww";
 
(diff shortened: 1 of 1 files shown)


LeoComandini:
This PR attempts to be as much consistent as possible with the existing code.

We could consider adding the token metadata with the same call that adds asset metadata and check consistency there.

src/pset/elip100.rs
```diff
@@ -46,13 +49,41 @@ impl PartiallySignedTransaction {
...
+        let key = prop_key(&asset_id, PSBT_ELEMENTS_HWW_GLOBAL_ASSET_METADATA);
 
         self.global
             .proprietary
             .get(&key)
             .map(|data| AssetMetadata::deserialize(data))
     }
+
+    /// Add token information to the PSET, returns None if it wasn't present or Some with the old
+    /// data if already in the PSET
```
RCasatta, line 61: Should we mention in the doc that you need also to add the relative contract metadata?

RCasatta: Or should we change `add_asset_metadata` to optionally accept `TokenMetadata`?

RCasatta (commented):
Concept ACK, checked test vectors matches the one in the elip doc

src/pset/elip100.rs
LeoComandini: Perhaps both.

In this PR I wanted to be the least disruptive possible, and just add a way to add/get token metadata in the PSET.

We could either:
* add an alternative method that sets asset and (optionally) token metadata at once (verifying their consistency and contract commitment)
* change `add_asset_metadata` to accept `Option<TokenMetadata>`

But those require a bit of discussion, and it might take time.

A viable approach could be to go with this PR and later add a "safer to use" method with a separate PR.
Then we can deprecate the `add/get_asset/token_metadata`, or keep them if we feel to. 
In this way we can think a bit more about the interface we want to expose, while not blocking who needs to use this new fields.

RCasatta:
ACK 11660892876f2e203db995ea93c651dd8273558a

PR #210, opened 2024-08-14
Address err refactor

RCasatta:
This is propedeutic to upgrade to bitcoin 0.32 #209 
where the base58 error variant specific to address are removed

Also it applies formatting around on which I think we agreed upon, but also on the toml file which I had doubt, let me know if I have to remove it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,32 +10,35 @@ documentation = "https://docs.rs/elements/"
 edition = "2018"
 
 [features]
-default = [ "json-contract" ]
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 42697283241bd251309a2d0ca3e1aadc2cc1e65d successfully ran local tests

apoelstra:
CI failures are on master, and appear to require we add some more pins (or fixed lockfiles) to CI. I'm fine to ignore it for now.

PR #212, opened 2024-08-19
Stop implementing elements::Encodable with bitcoin::Encodable

RCasatta:
Following bitcoin versions change `io` with `bitcoin_io`, upgrading would then require changing also elements::Encodable to match.

Instead, we are re-implementing what is needed.

Makes #209 easier

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -23,8 +23,8 @@ use std::io;
 use crate::dynafed;
 use crate::hashes::{Hash, sha256};
 use crate::Transaction;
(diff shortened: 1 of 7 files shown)


RCasatta:
I think this needs also to copy `ReadExt` and `WriteExt` from bitcoin...

apoelstra (approved):
ACK c08382ca1b5b42b022aff5fb83adda0c38e7423e successfully ran local tests; a bit redundant but I think this is the right way to go

PR #204, opened 2024-06-19
tx: add discount_weight and discount_vsize

delta1:
adds implementation of discount virtual size from ElementsProject/elements#1317 

same test vectors in https://github.com/ElementsProject/elements/pull/1341

PR #218, opened 2024-09-25
discount: fix weight calculation

delta1:
output amount and nonce discount must be weighted by the witness scaling factor as they form part of the base transaction, not the witness data 

see https://github.com/ElementsProject/elements/pull/1366


apoelstra (approved):
ACK 6909e1ccc1cdd19636bb4396f2b786f3b38012a8 successfully ran local tests; though I note that if we had a Weight type here like in rust-bitcoin we likely would have caught this :/

delta1:
link to #204 too

PR #221, opened 2025-02-12
ci: fixes for rust stable clippy, and rust 1.56.1 compilation

delta1:
Fixes clippy for the stable CI job, and pins libc to 0.2.163 for the rust 1.56.1 job

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -9,6 +9,7 @@ if cargo --version | grep "1\.56"; then
     cargo update -p which --precise 4.4.0
     cargo update -p byteorder --precise 1.4.3
     cargo update -p cc --precise 1.0.94
+    cargo update -p libc --precise 0.2.163
     cargo update -p serde_json --precise 1.0.98
     cargo update -p serde --precise 1.0.156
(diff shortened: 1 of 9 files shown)


RCasatta (approved):
ACK 01b0d2c9a2b4d21775cd9e0951dd29310949d54c

apoelstra:
Thanks for your patience -- I am struggling to get my local CI to work with this crate but I think I've almost got it. The issue is that our integration tests require Elements 0.21, which requires building with a very old version of gcc and boost, whose glibc is no longer compatible with the one included in nixos-unstable.

So when I went to build this crate with a new nixpkgs, and run the integration tests that used elementsd compiled with the old one, it wouldn't run (got some runtime missing symbol errors related to glibc 2_38 or something).

The fix was to bump my version gcc up to 13 (maaybe it's possible to just update glibc but use an old gcc, but I doubt it and I didn't try). To do this I had to use the new gcc13 patches in https://github.com/roconnor-blockstream/elements-nix/pull/4

Anyway, as a followup to this we should fix our integration tests to work with Elements 23, which builds with much newer tools (and should be more futureproof).

PR #224, opened 2025-03-17
pset: input: insert non-pset proprietary keys

LeoComandini:
As it is done for pset outputs.


apoelstra (approved):
ACK 45267c90e0694a2b97ba11b8c297e75178042528; successfully ran local tests

PR #229, opened 2025-06-28
Use bitcoin-units 1.0

apoelstra:
Draft until 1.0 is out; for now we are using rc0.

This introduces several changes that move this crate toward its final 1.0 form:

1. Remove some deprecated methods
2. Return the bitcoin-units `Weight` type from `weight` methods on `Transaction` and `Block`.
3. Replace ad-hoc serde impls on locktimes and PSETs with standard ones that use the consensus encoding.
4. Replace the `Height` and `Time` types in the locktime module with ones from bitcoin-units.

This will improve our compatibility with rust-bitcoin and also improve compatibility with different versions of rust-elements.


tcharding:
4-1 red lines to green, got to be  happy with that.

apoelstra:
Heh, well, much of it is replacing the serde stuff which I think I will PR separately because I want to get those changes in ASAP.

apoelstra:
Closing this. New strategy will be to use `bitcoin-consensus-encoding 1.0` first, and then to use the forthcoming compatibility layer in `bitcoin` 0.32.x.

PR #243, opened 2025-08-13
pset: blind: return ephemeral private key

LeoComandini:
Right now we have:
* `Transaction::blind` returns abf, vbf and ephemeral private key
* `PartiallySignedTransaction::blind_non_last` returns abf and vbf
* `PartiallySignedTransaction::blind_last` returns nothing

This PR makes things more consistent and makes ephemeral keys more easily accessible.

It's a breaking change, but it's just augmenting the data, so it should be easy to handle downstream.

### DIFF
--- a/src/blind.rs
+++ b/src/blind.rs
@@ -916,16 +916,29 @@ impl TxIn {
     }
 }
 
+/// Inputs or pseudo-inputs.
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum CtLocationType {
+    /// Regular input
+    Input,
+
+    /// Issuance pseudo-input
+    Issuance,
+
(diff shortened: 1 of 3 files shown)


apoelstra:
In a followup (or another commit, in case you get to this before I merge it), we should refactor `TxInType` to just be a flat enum, and have a separate struct with the index.

I think maybe

```
pub enum CtLocationType {
    Input,
    Issuance,
    Reissuance, // note I changed capitalization here
}

pub struct CtLocation {
    input_index: usize,
    ty: CtLocationType,
}
```

I'm not married to these names, but I think they're clearer than `TxInType`.

apoelstra (approved):
ACK e4144e6c5727c412ccf2b41f9372a1075a25beed; successfully ran local tests

PR #250, opened 2025-08-27
Release 0.26.1 with some minor API extensions

apoelstra:
I want to do the dtolnay "semver trick" where I release 0.25.3 which re-exports most of the types and traits from 0.26. This will make migration easier for users since they could use elements 0.25 and 0.26 in the same crate and have compatible types. In particular https://github.com/ElementsProject/elements-miniscript/pull/97 would compile without needing to update `simplicity` and `elements` at the same time.

However, when doing this I identified some gaps in the API which make this harder than it should be. See the commit messages for detailed justifications.

Also release 0.26.1 with these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 
+# 0.26.1 - 2025-08-28
+
+* [#250](https://github.com/ElementsProject/rust-elements/pull/250) API cleanups
+  * implement `Encodable` and `Decodable` for `Vec<T>` whenever `T` is `Encodable`/`Decodable` and `'static`
+  * add missing export of error sub-type `pset::PsetHash`
+
 # 0.26.0 - 2025-08-22
 
 * [#249](https://github.com/ElementsProject/rust-elements/pull/249) docs: fix changelog links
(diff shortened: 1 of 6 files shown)


apoelstra:
cc @RCasatta hopefully this isn't too bad to review. If you can ACK it I can merge it (and my merge script will cut the release automatically).

apoelstra (commented):
On 60de8df739469eae06dcdd2e36c13b606da50f83 successfully ran local tests

apoelstra:
Oh, yeah, I should add a changelog entry.

apoelstra:
Done. I just described the change. Saying "now you can implement this on your own types" is a little strong. It was always true that you could implement it on your own types... it's just that if you had a vector of those types you'd have an annoying time. (But not a *very* annoying time; you just need to define a newtype wrapper.)

apoelstra (commented):
On 05a24ad3715e8ba0649a2c756c906ff40e8b6f5c successfully ran local tests

apoelstra:
Tagged and published.

Issue #258, opened 2025-10-13
`std::error::Error` implementation for `hex::Error`


Velnbur:
It seems that `elements::hex::Error` is missing `std::error::Error` implementation, which makes it difficult to work with dynamic-error crates (like `eyre` or `anyhow`), as there default `::from` method requires that.

For example, 

```rust
let bytes_res = Vec::<u8>::from_hex(&raw_tx).map_err(eyre::Report::from);
```

doesn't compile, as `eyre::Report::from` requires `T: std::error::Error`, but:

```rust
let bytes_res = Vec::<u8>::from_hex(&raw_tx).map_err(|e| eyre::eyre!(e));
```

works, as it does some [black-magic](https://github.com/eyre-rs/eyre/blob/c6bf9c1f050971cdd0f94b46ce59f937da3e496b/eyre/src/error.rs#L96) under the hood with the value. So still `from` implementation is preferable

Velnbur:
@apoelstra  Added #259 check it out

PR #259, opened 2025-11-01
Replace custom hex parsing with hex-conservative crate 

Velnbur:
# Overview

Add hex-conservative as a dependency and refactor hex parsing code to use it instead of the custom implementation. Update FromHex trait to use associated error types and standardize hex decoding across the codebase.

Closes: #258

### DIFF
--- a/src/confidential.rs
+++ b/src/confidential.rs
@@ -720,6 +720,60 @@ impl<'de> Deserialize<'de> for Nonce {
     }
 }
 
+/// Error decoding hexadecimal string into tweak-like value.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TweakHexDecodeError {
+    /// Invalid hexadecimal string.
+    InvalidHex(hex_conservative::DecodeFixedLengthBytesError),
+    /// Invalid tweak after decoding hexadecimal string.
+    InvalidTweak(secp256k1_zkp::Error),
+}
+
+impl fmt::Display for TweakHexDecodeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            TweakHexDecodeError::InvalidHex(err) => {
+                write!(f, "Invalid hex: {}", err)
+            }
+            TweakHexDecodeError::InvalidTweak(err) => {
+                write!(f, "Invalid tweak: {}", err)
+            }
+        }
+    }
+}
+
+#[doc(hidden)]
+impl From<hex_conservative::DecodeFixedLengthBytesError> for TweakHexDecodeError {
+    fn from(err: hex_conservative::DecodeFixedLengthBytesError) -> Self {
+        TweakHexDecodeError::InvalidHex(err)
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/encode.rs
```diff
@@ -54,8 +55,10 @@ pub enum Error {
     Secp256k1zkp(secp256k1_zkp::Error),
     /// Pset related Errors
     PsetError(pset::Error),
-    /// Hex parsing errors
-    HexError(crate::hex::Error),
+    /// Hex fixed parsing errors
+    HexFixedError(DecodeFixedLengthBytesError),
+    /// Hex variable parsing errors
+    HexVariableError(DecodeVariableLengthBytesError),
```
Velnbur: Not really sure if it's okay to keep both this way. We can alternatively keep only one variant as before `HexError()`, and add `hex::Error` which is merge of `DecodeFixedLengthBytesError` and `DecodeVariableLengthBytesError` and add `From` implementations

apoelstra:
> What made you think so? I made this changes without any AI tools. Except commit messages, I use automation tools to write them, shouldn't be a sin nowadays . May be some words like "for more robust hexadecimal encoding/decoding" sound AI-ish, kind of understand that

I apologize -- and for what it's worth, I don't mind the use of AI tools or even majority-LLM-authored PRs. What's a problem is that Github is now full of unsupervised agents which open broken PRs and then respond to any comments or requests by closing their PR and moving on. It's extremely irksome to be scammed by a machine into doing pointless work. (There are some technical limitations keeping us on Github but I hope to abandon it entirely within the next year or two because of this.)

For future reference -- what triggered me was mostly the "Overview" heading in your PR comment, which looks like LLM output (humans don't usually write PR descriptions in structured markdown), combined with the unit tests not passing (though humans do this all the time :P).

Velnbur:
@delta1 Check please latest commits, I added `Cargo-latest.lock` and tests are passing (at least those that start with `cargo test`) as you asked. Am I missing something? My apologies if so

src/confidential.rs
```diff
@@ -720,6 +720,60 @@ impl<'de> Deserialize<'de> for Nonce {
...
+            TweakHexDecodeError::InvalidHex(err) => encode::Error::HexFixedError(err),
+            TweakHexDecodeError::InvalidTweak(err) => encode::Error::Secp256k1zkp(err),
+        }
+    }
+}
+
+impl std::error::Error for TweakHexDecodeError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            TweakHexDecodeError::InvalidHex(err) => Some(err),
```
apoelstra: In 775d7cd883dd5e45f059cb7fec73a6fa2fdbfe9b:

Also don't need to fix it here but we should use `Self::` instead of writing `TweakHexDecodeError::` all over the place.

apoelstra (approved):
ACK 775d7cd883dd5e45f059cb7fec73a6fa2fdbfe9b; successfully ran local tests

PR #261, opened 2025-12-24
Removed MAX_MONEY

Arvolear:
Removed `MAX_MONEY` restriction as it is only useful for Bitcoin. As discussed with Andrew, `rest-elements` should be chain-agnostic, so no need for this check to exist.

I believe this restriction may also be relaxed in the `elements` repository for alternative assets. By doing so, it would allow more useful information to be stored on-chain (up to 12 more bits per UTXO!).

### DIFF
--- a/src/blind.rs
+++ b/src/blind.rs
@@ -39,8 +39,6 @@ pub enum TxOutError {
     UnExpectedNullValue,
     /// Unexpected Null asset
     UnExpectedNullAsset,
-    /// Money should be between 0 and `21_000_000`
-    MoneyOutofRange,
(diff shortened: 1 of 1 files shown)


apoelstra:
Please remove e8cff54bf25a9168928ebaa2f7ba362f7b88f9e3 from this PR.

Arvolear:
Removed the commit.

Oh, great to see an ELIP regarding this topic. Will check it out.

apoelstra (approved):
ACK 33d4b17f45a68eb71ceb3564a7ea9149260eccec; successfully ran local tests

PR #257, opened 2025-10-06
Automated daily update to rustc (to nightly-2026-04-18)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #266, opened 2026-04-21
Update cargo-rbmt and fix CI

apoelstra:
This overhauls the CI system (again) by copying all the files from rust-bech32 and rust-bitcoin and making minor tweaks for rust-elements.

This should fix the `doc_auto_cfg` thing which is plaguing #257, and should overall be much more reliable going forward since it's much more agressive about pinning toolchain versions etc.

This does not re-enable the fuzz tests (which were not covered by the previous CI and not covered by this one); do this will require I migrate the old honggfuzz-based fuzz targets to cargo-fuzz ones, and require I migrate the ancient travis-fuzz.sh script to our more modern one. Will do this in a separate PR.


apoelstra:
cc @delta1 can you take a look at this? A cursory review is fine, this stuff is CI only.

apoelstra:
On 89f15f4ba932330a178ab1a4c7919daec829fed5  successfully ran local tests

delta1 (approved):
ACK 89f15f4ba932330a178ab1a4c7919daec829fed5; ran tests locally 

Thanks for fixing the CI @apoelstra !

PR #267, opened 2026-04-22
Get fuzztests running daily in CI

apoelstra:
Copied from rust-bitcoin: run the fuzztests every day in Github CI, file issues when they fail.

Convert the fuzztests from honggfuzz to cargo-fuzz, which is (maybe) slower but much more reliable.


apoelstra:
My local CI does not like this because libfuzzer-sys won't build on 1.63 due to MSRV. Meanwhile, I cannot bump the MSRV of `elements-fuzz` separate from the wider MSRV because `cargo-rbmt` rejects it.

I propose we just bump the MSRV of this whole crate to 1.74 to match the rust-bitcoin ecosystem.

apoelstra:
Blocked on #269.

delta1:
Merged #269

apoelstra:
On 20440d3c6229210cab8ac9dcd45f872362d76f15 successfully ran local tests

delta1 (approved):
ACK 20440d3c6229210cab8ac9dcd45f872362d76f15; ran tests locally

PR #269, opened 2026-04-22
bump MSRV to 1.74.0

apoelstra:
This matches the rust-bitcoin ecosystem, and in particular bitcoin-consensus-encoding which we hope to replace our homebrew encoding traits with. This MSRV is also needed to use libfuzzer, which is used by cargo-fuzz.

lwk and rust-simplicity both have higher MSRVs so there seems to be little point here in having a very old one.


apoelstra:
ACK a73585414a47faf5eacb6a17411260d6b8756362; successfully ran local tests

delta1 (approved):
ACK a73585414a47faf5eacb6a17411260d6b8756362; tested locally

PR #272, opened 2026-05-09
Update to hex-conservative 1.1; drop `hex` module

apoelstra:
We have a `hex` module which is loosely based on the `hashes::hex` module from `bitcoin_hashes` 0.11 or so. rust-bitcoin has updated to the `hex-conservative` crate, "hex with a conservative MSRV policy", which is maintained by the rust-bitcoin community.

We recently have released version 1.1, which is finally feature-complete, after years of haggling over API questions.

This has a couple philosophical changes from the old `hex` module:

* There is no `FromHex` trait; the assumption is that if something is naturally hex-encoded then it should use `FromStr` to parse from hex, and there is little value in a trait describing "an object which can be parsed from hex" beyond this. For things like `Script` which have a hex-encoding but aren't "naturally" hex-encoded, they should have inherent methods
* Relatedly, rather than requiring extension traits to decode hex to arrays or vectors, we provide standalone methods `decode_to_vec` and `decode_to_array`
* The `ToHex` trait, which previously was oriented around producing hex-encoded `String`s, has been replaced with `DisplayHex`, a trait which produces a (typically) no-alloc wrapper around an object that can be hex-formatted. It is a trait rather than standalone function(s) so it can be cleanly implemented for both `[u8]` and `[u8]`-like things, but we don't expect users to implement this; they should use `Display` or inherent methods instead

In addition, there are a pile of mechanical changes; `to_hex` becomes `as_hex` in most cases, `Vec::from_hex` becomes `decode_to_vec` or `hex!`, etc. So this PR has a large diff but it's mostly mechanical, and the changes serve both to clean up the code and to eliminate a bunch of allocations.

A later PR will replace our encoding traits with the new `bitcoin-consensus-encoding` trait, which will further eliminate allocations when deserializing objects from hex.


apoelstra:
ACK 44a6974082031652ead21cfb1d88d09a9b8fb8da; successfully ran local tests

delta1 (approved):
ACK 44a6974082031652ead21cfb1d88d09a9b8fb8da; tested locally

PR #276, opened 2026-06-08
Implement calculation of AssetId and genesis blocks

philipr-za:
Recently had cause to work with custom elements chains in regtest mode and needed some helpers to calculate the pegged in AssetId and genesis block hash for those custom network parameters.

This PR implements these helpers based on how it is done in `elementsd`

### DIFF
--- a/src/issuance.rs
+++ b/src/issuance.rs
@@ -20,7 +20,9 @@ use crate::encode::{self, Encodable, Decodable};
 use crate::hashes::{hash_newtype, sha256, sha256d, Hash};
 use crate::fast_merkle_root::fast_merkle_root;
 use secp256k1_zkp::Tag;
+use crate::genesis::{commit_to_custom_network_parameters, NetworkParams};
 use crate::transaction::OutPoint;
+use crate::Txid;
 
 /// The zero hash.
 const ZERO32: [u8; 32] = [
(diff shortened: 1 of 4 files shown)


src/issuance.rs
```diff
@@ -75,8 +76,16 @@ impl AssetId {
...
+    pub const LIQUIDTESTNET_BTC: AssetId = AssetId(sha256::Midstate([
+        0x49, 0x9a, 0x81, 0x85, 0x45, 0xf6, 0xba, 0xe3,
+        0x9f, 0xc0, 0x3b, 0x63, 0x7f, 0x2a, 0x4e, 0x1e,
+        0x64, 0xe5, 0x90, 0xca, 0xc1, 0xbc, 0x3a, 0x6f,
+        0x6d, 0x71, 0xaa, 0x44, 0x43, 0x65, 0x4c, 0x14,
+    ]));
+
     /// Create an [`AssetId`] from its inner type.
-    pub const fn from_inner(midstate: sha256::Midstate) -> AssetId {
+    pub fn from_inner(midstate: sha256::Midstate) -> AssetId {
```
delta1: Is this API breaking change necessary? I think it should still work `const`?

src/issuance.rs
```diff
@@ -146,6 +155,31 @@ impl AssetId {
...
+            _ => {
+                // Else calculate the asset_id
+                let asset_id = Self::pegged_asset_id_for_params_and_parent_chain_hash(
+                    params,
+                    bitcoin::Network::Regtest.chain_hash(),
+                );
+                Some(asset_id)
+            }
+        }
+    }
```
delta1: No branch returns `None`, is it necessary that this function returns an `Option`?

delta1 (commented):
Thanks @philipr-za this looks great, just two review comments

src/issuance.rs
philipr-za: Indeed it does, not sure why I did that. fixed.

src/issuance.rs
philipr-za: Nice spot, this branch came from a larger batch of work where I had some other options but making the scope smaller means it is no longer needed and I missed fixing this method.

Fixed.

delta1 (approved):
ACK 437f1761cf606a45d82380eadd91f9ed95661bbe; tested locally

apoelstra:
In 1d9111b82df81ccdbb3546c5bd0c4c69b8dafb74:

```
error[E0432]: unresolved import `crate::genesis`
  --> src/issuance.rs:23:12
   |
23 | use crate::genesis::{commit_to_custom_network_parameters, NetworkParams};
   |            ^^^^^^^ could not find `genesis` in the crate root

error: aborting due to previous error
```

The tip commit is fine so I think there are just some changes in the wrong commit. Other than this, ack.

philipr-za:
Thanks, rejiggered the commits to compile cleanly at each stage

apoelstra:
9b0e695a9b05332f261fc374a7d6d78addbf6d27 needs rebase

apoelstra (approved):
ACK d7f8720f436164c703b58d9ba5f6315c48a8a801; successfully ran local tests

PR #278, opened 2026-06-12
Clean up hashes and related types ahead of `bitcoin_hashes` 1.0

apoelstra:
I am going to PR soon to update to bitcoin-hashes 1.0. There are several big changes versus the old version of bitcoin-hashes that we were using:

* Hash wrapper types (like `Txid`, `Blockhash`, etc) no longer have general hashing methods like `hash` that let you hash up arbitrary data; the intention of these types is that they only be constructable from byte slices or by hashing the right kind of data
* We no longer directly support construction from byte slices; instead you must use arrays. stdlib has a `TryFrom<&[u8; N]> for &[u8]` impl which is just a pointer cast, so you can get the old behavior by using this `TryFrom`
* The `sha256::Midstate` type now carries a length with it, so it behaves something like a deconstructed hash engine rather than being a "hash" which is computed from the sha256 compression function; you can no longer effectively wrap a midstate the way you would wrap a sha256 hash, so instead we directly wrap `[u8; 32]`s for the types that did this
* Many changes to make the macros more ergonomic, standard trait impls more consistent and complete, etc

This PR makes a bunch of prepatory changes to make this transition smaller. In particular, it

* eliminate all the bare `sha256::Midstate` uses by introducing newtypes for each midstate (asset entropy, dynafed elided params, etc)
* replace slices with arrays in many places throughout the codebase, eliminating a ton of unreachable panic paths
* **makes `Address::p2wpkh` and `Address::p2shwpkh` panic if you give them uncompressed keys** rather than producing unspendable addresses; later we will clean this up further by type-separating compressed keys, but for now ISTM that a panic is better than silently creating black-hole addresses
* does a whole pile of code cleanups

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,10 @@ default = ["json-contract"]
 
 json-contract = ["serde_json"]
 "serde" = [
+    "dep:serde",
     "bitcoin/serde",
     "bitcoin/serde",
     "secp256k1-zkp/serde",
-    "actual-serde",
 ]
 base64 = ["bitcoin/base64"]
 
@@ -33,10 +33,7 @@ secp256k1-zkp = { version = "0.11.0", features = ["global-context", "hashes"] }
 
 # Used for ContractHash::from_json_contract.
 serde_json = { version = "1.0", optional = true }
-
-actual-serde = { package = "serde", version = "1.0.103", features = [
-    "derive",
-], optional = true }
+serde = { version = "1.0.103", features = [ "derive" ], optional = true }
 hex = { package = "hex-conservative", version = "1.1.0" }
 
 
--- a/src/address.rs
+++ b/src/address.rs
@@ -36,7 +36,7 @@ use crate::schnorr::{TapTweak, TweakedPublicKey, UntweakedPublicKey};
 use crate::taproot::TapNodeHash;
 
 use crate::{opcodes, script};
-use crate::{PubkeyHash, ScriptHash, WPubkeyHash, WScriptHash};
+use crate::{PubkeyHash, ScriptHash, WScriptHash};
 
 /// Encoding error
 #[derive(Debug, PartialEq)]
@@ -236,13 +236,9 @@ impl Address {
         blinder: Option<secp256k1_zkp::PublicKey>,
         params: &'static AddressParams,
     ) -> Address {
-        let mut hash_engine = PubkeyHash::engine();
-        pk.write_into(&mut hash_engine)
-            .expect("engines don't error");
-
         Address {
             params,
-            payload: Payload::PubkeyHash(PubkeyHash::from_engine(hash_engine)),
+            payload: Payload::PubkeyHash(pk.pubkey_hash()),
             blinding_pubkey: blinder,
         }
     }
@@ -264,20 +260,20 @@ impl Address {
 
     /// Create a witness pay to public key address from a public key
     /// This is the native segwit address type for an output redeemable with a single signature
+    ///
+    /// # Panics
+    ///
+    /// Panics if the provided public key is not compressed.
     pub fn p2wpkh(
         pk: &PublicKey,
         blinder: Option<secp256k1_zkp::PublicKey>,
         params: &'static AddressParams,
     ) -> Address {
-        let mut hash_engine = WPubkeyHash::engine();
-        pk.write_into(&mut hash_engine)
-            .expect("engines don't error");
-
         Address {
             params,
             payload: Payload::WitnessProgram {
                 version: Fe32::Q,
-                program: WPubkeyHash::from_engine(hash_engine)[..].to_vec(),
+                program: pk.wpubkey_hash().expect("public key must be compressed").as_byte_array().to_vec(),
             },
             blinding_pubkey: blinder,
         }
--- a/src/address.rs
+++ b/src/address.rs
@@ -253,7 +253,7 @@ impl Address {
     ) -> Address {
         Address {
             params,
-            payload: Payload::ScriptHash(ScriptHash::hash(&script[..])),
+            payload: Payload::ScriptHash(ScriptHash::hash_script(script)),
             blinding_pubkey: blinder,
         }
     }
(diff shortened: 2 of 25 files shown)


apoelstra:
On f6d247ea98b0eb199756b7490abd0a6add12c261 successfully ran local tests

src/lib.rs
```diff
@@ -34,8 +34,7 @@ pub extern crate hex;
 pub extern crate secp256k1_zkp;
 /// Re-export of serde crate
 #[cfg(feature = "serde")]
-#[macro_use]
-pub extern crate actual_serde as serde;
+pub extern crate serde;
```
stringhandler, line 37: nit: probably doesn't need to be `pub`. If the users want serde, they will probably reference it anyway

Cargo.toml
```diff
@@ -33,10 +33,7 @@ secp256k1-zkp = { version = "0.11.0", features = ["global-context", "hashes"] }
 
 # Used for ContractHash::from_json_contract.
 serde_json = { version = "1.0", optional = true }
-
-actual-serde = { package = "serde", version = "1.0.103", features = [
-    "derive",
-], optional = true }
+serde = { version = "1.0.103", features = [ "derive" ], optional = true }
```
stringhandler, line 36: nit: This can probably just be `1` as it's a library and we want to specify lowest compatible version

stringhandler (commented):
utACK dbfbaf8b22a2f6ebf7f6c5294609352e92a06bb4

stringhandler (commented):
utACK dc701a5976ec895b46f3e0dca85c53ee1c9a96a0

src/lib.rs
apoelstra: For `serde` this is probably true because there is only one major version of serde that anybody will ever use.

But as a matter of policy we like to do `pub extern crate` so that users can easily get "the version of the dependency used by rust-elements".

Cargo.toml
apoelstra: That would mean 1.0.0, which doesn't even compile on modern Rust compilers. 1.0.103 is the minimum version that I've tested to be compatible. (Though it actually looks suspiciously low ... on rust-bitcoin we have been using 1.0.158 or something for a while.)

But I think cleaning up the serde dep is out of scope for this PR.

stringhandler (commented):
utACK 0c006d721efcfec47ee41975e1d1072a9012e9ec

src/lib.rs
stringhandler: noted

Cargo.toml
stringhandler: noted.

stringhandler (commented):
utACK da967f81a4cb0e891a5e2a1454eb209ed4bbef03

One thing to note is that the error returned has changed, but I couldn't see any checks for the Pset error in particular

stringhandler (commented):
utACK 7ec80feae0e54bb1048e3354d3d7dc595ba94e4e

stringhandler (commented):
utACK 90183f4b55a3c4ffafab671c09800467488df7d4

stringhandler (commented):
utACK 5f54fefd03bf749fef2df7b55fd8a428c34dfaa6

stringhandler (commented):
utACK da8942ac3fd87c1bb16cfb2af41e2006a7dde5a9

stringhandler (commented):
utACK b985342ef5126a1dffb67be0980239031c0e68d1

stringhandler (commented):
utACK 8153c4a88c91926a9aaa52c12b16ce99a18195dd

stringhandler (commented):
utACK 5c226590b9c7f0d334b9bd0b30f57ef3c48ce026

src/address.rs
```diff
@@ -264,20 +260,20 @@ impl Address {
...
-        let mut hash_engine = WPubkeyHash::engine();
-        pk.write_into(&mut hash_engine)
-            .expect("engines don't error");
-
         Address {
             params,
             payload: Payload::WitnessProgram {
                 version: Fe32::Q,
-                program: WPubkeyHash::from_engine(hash_engine)[..].to_vec(),
+                program: pk.wpubkey_hash().expect("public key must be uncompressed").as_byte_array().to_vec(),
```
stringhandler: In fe59e1a728f1714a17b19d2693c0768478579d46

Did you mean "must be compressed"? Otherwise the comment above it wrong.

stringhandler (commented):
utACK fe59e1a728f1714a17b19d2693c0768478579d46

The panic comment is potentially wrong though

src/address.rs
```diff
@@ -281,22 +281,23 @@ impl Address {
...
     pub fn p2shwpkh(
         pk: &PublicKey,
         blinder: Option<secp256k1_zkp::PublicKey>,
         params: &'static AddressParams,
     ) -> Address {
-        let mut hash_engine = ScriptHash::engine();
-        pk.write_into(&mut hash_engine)
-            .expect("engines don't error");
-
+        let pkh = pk.wpubkey_hash().expect("public key must be uncompressed");
```
stringhandler: In f6d247ea98b0eb199756b7490abd0a6add12c261

This should either be "compressed" or the doc comment be "compressed"

stringhandler (commented):
In f6d247ea98b0eb199756b7490abd0a6add12c261

Unfortunately I'm not familiar enough with this code to know if it's correct or not. The one comment does seem off though

src/address.rs
apoelstra: oops, yes, I mean "must be compressed"

stringhandler (approved):
utACK b553acb97af691b7568071b66beedd2b3ccf98b5

src/address.rs
stringhandler: This one still remains @apoelstra

apoelstra:
On 577e503250a7f6579c328d69ff5181fab83c1609 successfully ran local tests

stringhandler (approved):
utACK b553acb97af691b7568071b66beedd2b3ccf98b5

stringhandler (approved):
utACK fa3ea85ff7968412e4532119233e87d52a4a9870

apoelstra:
On fa3ea85ff7968412e4532119233e87d52a4a9870 successfully ran local tests

PR #279, opened 2026-06-15
Improve slicing and error handling around slices

apoelstra:
Uses the `bitcoin-internals` library which provides us a bunch of array/slice methods that aren't available in std in our MSRV. Eventually we will be able to drop the dep and get these features directly from std.

This replaces a bunch of slice accesses with arrays, cleaning up error paths and eliminating panics (though most are inaccessible assuming you use known-valid data from the blockchain).

Builds on #278 (which I will merge in the next hour).

Will do a followup that cleans up the error types; I deliberately avoided breaking the API to make this one easy to backport.

### DIFF
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -26,7 +26,7 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
 dependencies = [
- "bitcoin-internals",
+ "bitcoin-internals 0.3.0",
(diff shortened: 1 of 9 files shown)


philipr-za:
ACK 5e0d5774d96313e3d65dc47041cedc907a40b8de 

The type inference that auto-magically figures out the sizes for `split_array` is neat!

apoelstra:
Yeah, these are great APIs and frustratingly they have been unstable for a long time and still not in stable https://doc.rust-lang.org/std/primitive.array.html#method.split_array_ref

apoelstra:
ACK 5e0d5774d96313e3d65dc47041cedc907a40b8de; successfully ran local tests

PR #280, opened 2026-06-15
Backport slicing improvements to 0.25.x

philipr-za:
There are a number of useful improvements from https://github.com/ElementsProject/rust-elements/pull/279 we would like to include in 0.25.x.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,7 +84,7 @@ jobs:
         env:
           DO_DOCS: false
           DO_DOCSRS: false
-          DO_FUZZ: true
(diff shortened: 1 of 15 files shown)


apoelstra:
In 33ca246a128572b87f6bad8a62b26f688ea610be:

This violates the MSRV.

philipr-za:
Indeed, I only needed one helper from 278 actually so manually added that and dropped the rest of 278

apoelstra:
utACK 982a50f599e72b72c95908f72d4a88d65793f7aa

apoelstra:
Tagged and published.

PR #286, opened 2026-06-30
refactor confidential.rs and clean up code

apoelstra:
This is a lot of commits but hopefully should be straightforward to review. Two are formatting, one is a rename-only, at least one is code-move-only, and a few others just enable lints. The big changes are:

1. We encapsulate `RangeProof` and `SurjectionProof` into dedicated types, rather than using `Option<Box<secp256k1_zkp::RangeProof>>` everywhere. This fixes a type confusion in the PSET code and overall greatly simplifies all our interactions with rangeproofs and surjection proofs.
2. We break apart the three confidential commitment types into their own modules and refactor the code a bunch to try to minimize the text-diff between the modules. This hopefully will prevent divergence between these three basically-the-same objects. But there is no user-visible effect to this change.
3. We remove `Encodable/Decodable` for secp256k1_zkp objects, since these should never appear "bare" in rust-elements code. Instead we implement these only on rust-elements wrapper types. We also simplify these implementations a bit to e.g. directly encode big-endian integers for confidential values rather than calling `u64::consensus_encode` then `u64::swap_bytes`.

The next PR will introduce the rust-bitcoin 0.32.100 series, which have new `Encode`/`Decode` traits with much better error handling, no more `bitcoin-io`, and fewer allocations. This PR helps set the stage for that by cleaning up the existing `Encodable/Decodable` impls and improving the overall type safety of the library.


apoelstra:
On cf24680388f86cd0a97d3a3fc9503a108645212c successfully ran local tests

src/pset/map/output.rs
```diff
@@ -233,8 +233,8 @@ impl Output {
             });
         }
         rv.script_pubkey = txout.script_pubkey;
-        rv.value_rangeproof = txout.witness.rangeproof;
-        rv.asset_surjection_proof = txout.witness.surjection_proof;
+        rv.value_rangeproof = Some(txout.witness.rangeproof);
+        rv.asset_surjection_proof = Some(txout.witness.surjection_proof);
```
delta1: Previously we did direct assignment (`rv.value_rangeproof = txout.witness.rangeproof;`) on `Option<Box<RangeProof>>`, so `None` propagated for any unblinded output/issuance.  

Every PSET built from an ordinary (non-confidential) transaction now emits spurious zero-length `PSBT_ELEMENTS_OUT_VALUE_RANGEPROOF/_ASSET_SURJECTION_PROOF/_ISSUANCE_*_RANGEPROOF` keys that master correctly omitted. This also corrupts `is_partially_blinded()`/`is_fully_blinded()`, which key off `Option::is_some()` on these same fields — an output can now spuriously appear "partially blinded" and trip `Error::MissingBlindingInfo` on decode when it shouldn't. 

Here's a failing unit test that repros in `src/pset/mod.rs`: 

```rust

    /// `Output::from_txout` (and the analogous `Input::from_txin`/blinding-path
    /// code) must not wrap `RangeProof`/`SurjectionProof` in `Some(...)` for a
    /// fully explicit (unblinded) transaction where no proof exists. Doing so
    /// loses the `None`/"absent" signal that `Option<Box<T>>` used to carry, so:
    ///   1. `is_partially_blinded()`/`is_fully_blinded()` (which key off
    ///      `Option::is_some()`) would misreport an unblinded output as
    ///      carrying proof data, and
    ///   2. the PSET serializer would emit a spurious zero-length proprietary
    ///      key/value pair for every unblinded output/issuance that should be
    ///      omitted entirely.
    #[test]
    fn from_tx_does_not_spuriously_set_proofs_on_unblinded_outputs() {
        // Same fully-explicit (no confidential value/asset, no issuance)
        // transaction used by `test_pset` below.
        let tx_hex = "010000000001715df5ccebaf02ff18d6fae7263fa69fed5de59c900f4749556eba41bc7bf2af0000000000000000000201230f4f5d4b7c6fa845806ee4f67713459e1b69e8e60fcee2e4940c7a0d5de1b2010000000124101100001f5175517551755175517551755175517551755175517551755175517551755101230f4f5d4b7c6fa845806ee4f67713459e1b69e8e60fcee2e4940c7a0d5de1b2010000000005f5e100000000000000";
        let tx: Transaction =
            encode::deserialize(&hex::decode_to_vec(tx_hex).unwrap()).unwrap();
        assert!(tx.output.iter().all(|o| o.witness.is_empty()), "fixture must be unblinded");

        let pset = PartiallySignedTransaction::from_tx(tx);

        for (i, out) in pset.outputs().iter().enumerate() {
            assert!(
                out.value_rangeproof.is_none(),
                "output {i}: value_rangeproof should be None for an unblinded output, got {:?}",
                out.value_rangeproof,
            );
            assert!(
                out.asset_surjection_proof.is_none(),
                "output {i}: asset_surjection_proof should be None for an unblinded output, got {:?}",
                out.asset_surjection_proof,
            );
        }

        // Wire-format check: an unblinded output's key-value pairs must not
        // include the value-rangeproof / asset-surjection-proof proprietary
        // keys at all (subtypes 0x04/0x05 per PSBT_ELEMENTS_OUT_VALUE_RANGEPROOF/
        // PSBT_ELEMENTS_OUT_ASSET_SURJECTION_PROOF in pset/map/output.rs), not
        // even with an empty value. `Map::get_pairs` is exactly what
        // `consensus_encode` iterates over (see `impl_pset_consensus_encoding!`
        // in pset/macros.rs), so this checks the real wire format rather than
        // grepping serialized bytes.
        let value_rangeproof_key = raw::ProprietaryKey::from_pset_pair(0x04, vec![]).to_key();
        let asset_surjection_proof_key = raw::ProprietaryKey::from_pset_pair(0x05, vec![]).to_key();
        for (i, out) in pset.outputs().iter().enumerate() {
            let pairs = Map::get_pairs(out).unwrap();
            assert!(
                !pairs.iter().any(|p| p.key == value_rangeproof_key),
                "output {i}: serialized pairs unexpectedly contain a value-rangeproof key: {:?}",
                pairs,
            );
            assert!(
                !pairs.iter().any(|p| p.key == asset_surjection_proof_key),
                "output {i}: serialized pairs unexpectedly contain an asset-surjection-proof key: {:?}",
                pairs,
            );
        }
    }
```

src/pset/serialize.rs
```diff
@@ -282,31 +282,27 @@ impl Deserialize for confidential::Asset {
...
     }
 }
 
-impl Deserialize for Box<RangeProof> {
+impl Deserialize for RangeProof {
     fn deserialize(bytes: &[u8]) -> Result<Self, encode::Error> {
-        let prf = RangeProof::from_slice(bytes)
-            .map_err(|_| encode::Error::ParseFailed("Invalid Rangeproof"))?;
-        Ok(Box::new(prf))
+        Self::from_slice(bytes).map_err(encode::Error::Secp256k1zkp)
```
delta1, line 293: A PSET with a deliberately zero-length rangeproof/surjection-proof value is now accepted during parsing where it previously would have been rejected. 

```rust
    /// `RangeProof`/`SurjectionProof`'s
    /// low-level `from_slice` special-cases an empty byte slice as
    /// `Self::EMPTY` (see `src/confidential/range_proof.rs`/`surjection_proof.rs`).
    /// The pset-specific `serialize::Deserialize` impls for these types
    /// (`src/pset/serialize.rs`) reuse that same `from_slice`, so decoding a
    /// zero-length PSET proprietary value now silently succeeds as `EMPTY`.
    /// Previously (on master), `Deserialize for Box<RangeProof>`/
    /// `Box<SurjectionProof>` called straight through to
    /// `secp256k1_zkp::RangeProof::from_slice`/`SurjectionProof::from_slice`,
    /// which reject empty input, so this decode used to fail closed with
    /// `Err(ParseFailed("Invalid Rangeproof"/"Invalid SurjectionProof"))`.
    /// A PSET crafted with a zero-length value for one of these proprietary
    /// keys is now accepted where it previously would have been rejected.
    #[test]
    fn pset_deserialize_rejects_empty_rangeproof_and_surjectionproof_bytes() {
        use crate::pset::serialize::Deserialize as PsetDeserialize;

        let rangeproof_result = <RangeProof as PsetDeserialize>::deserialize(&[]);
        assert!(
            rangeproof_result.is_err(),
            "deserializing an empty byte slice as a RangeProof should fail (no valid \
             rangeproof is zero bytes), got {:?}",
            rangeproof_result,
        );

        let surjectionproof_result = <SurjectionProof as PsetDeserialize>::deserialize(&[]);
        assert!(
            surjectionproof_result.is_err(),
            "deserializing an empty byte slice as a SurjectionProof should fail (no valid \
             surjection proof is zero bytes), got {:?}",
            surjectionproof_result,
        );
    }
```

delta1 (commented):
This generally looks great! But I left comments with some issues found with Claude's help

src/pset/map/output.rs
apoelstra: > Previously we did direct assignment (`rv.value_rangeproof = txout.witness.rangeproof;`) on `Option<Box<RangeProof>>`, so `None` propagated for any unblinded output/issuance.

`None` was encoded as an empty rangeproof. There was no possible way to express "no rangeproof present in the PSET".

But you're right -- I misread this code as extracting a `TxOut` from a `pset::Output` but it's the other way around. Will fix (and take your unit test).

src/pset/serialize.rs
apoelstra: >A PSET with a deliberately zero-length rangeproof/surjection-proof value is now accepted during parsing where it previously would have been rejected.

This check seems to be only for the purpose of catching mistakes. It never checked whether the user put an invalid rangeproof into the PSET, only whether the rangeproof was present or not.

So it seems to me that a deliberately-insert 0-length rangeproof is no different than any other kind of deliberately-insert invalid rangeproof. But I could be convinced otherwise.

src/pset/map/output.rs
apoelstra: Added a commit with this unit test and fixed the two offending commits which caused it to fail.

delta1: Thanks! Sorry I forgot to mention that `Input::from_txin` also needs a similar fix

src/pset/map/input.rs
```diff
@@ -540,8 +540,8 @@ impl Input {
             }
 
             // Witness
-            ret.issuance_keys_rangeproof = txin.witness.inflation_keys_rangeproof;
-            ret.issuance_value_rangeproof = txin.witness.amount_rangeproof;
+            ret.issuance_keys_rangeproof = Some(txin.witness.inflation_keys_rangeproof);
+            ret.issuance_value_rangeproof = Some(txin.witness.amount_rangeproof);
```
delta1: This needs the same treatment, then should be good to go

apoelstra: Can you claude-generate me a unit test? (Or maybe we just shouldn't bother?)

src/pset/map/output.rs
apoelstra: Fixed. (But did not add a unit test.)

src/pset/map/input.rs
delta1: Sure will do

delta1:
```rust
#[cfg(test)]
mod tests {
    use crate::pset::PartiallySignedTransaction;
    use crate::{
        confidential, AssetIssuance, LockTime, Sequence, Transaction, TxIn, TxInWitness,
    };
    use secp256k1_zkp::ZERO_TWEAK;

    // Sibling of `pset::map::output::tests::
    // from_tx_does_not_spuriously_set_proofs_on_unblinded_outputs`, but for
    // an input with an *explicit* (unblinded) issuance: `Input::from_txin`
    // still unconditionally wraps `issuance_value_rangeproof`/
    // `issuance_keys_rangeproof` in `Some(...)`, even though an explicit
    // issuance has no rangeproof at all (both are `RangeProof::EMPTY`).
    #[test]
    fn from_tx_does_not_spuriously_set_proofs_on_explicit_issuance() {
        let txin = TxIn {
            asset_issuance: AssetIssuance {
                asset_blinding_nonce: ZERO_TWEAK,
                asset_entropy: [1u8; 32],
                amount: confidential::Value::Explicit(1000),
                inflation_keys: confidential::Value::Null,
            },
            witness: TxInWitness::empty(),
            ..TxIn::default()
        };
        assert!(txin.has_issuance());
        assert!(txin.witness.amount_rangeproof.is_empty());
        assert!(txin.witness.inflation_keys_rangeproof.is_empty());

        let tx = Transaction {
            version: 2,
            lock_time: LockTime::ZERO,
            input: vec![txin],
            output: vec![],
        };

        let pset = PartiallySignedTransaction::from_tx(tx);
        let input = &pset.inputs()[0];
        assert!(
            input.issuance_value_rangeproof.is_none(),
            "issuance_value_rangeproof should be None for an explicit issuance, got {:?}",
            input.issuance_value_rangeproof,
        );
        assert!(
            input.issuance_keys_rangeproof.is_none(),
            "issuance_keys_rangeproof should be None for an explicit issuance, got {:?}",
            input.issuance_keys_rangeproof,
        );
    }
}
```

apoelstra:
Added the new unit test.

apoelstra:
On e9ffc698a2d966b24bd82d95957024f4b576b5e5 successfully ran local tests

delta1 (approved):
ACK e9ffc698a2d966b24bd82d95957024f4b576b5e5; tested locally

PR #288, opened 2026-07-15
Introduce `bitcoin-consensus-encoding 1.0` and implement traits for `Transaction` and subtypes

apoelstra:
This is a big PR. Please forgive me.

The goal is to replace the `Encodable` and `Decodable` traits from rust-elements with the new [`Encode` and `Decode` traits from bitcoin-consensus-encoding](https://docs.rs/bitcoin-consensus-encoding/latest/bitcoin_consensus_encoding/index.html). There are many benefits to this:

* Unlike the old rust-bitcoin traits, these ones are designed to be used by the general public for wire encoding formats, so we don't need to create our own duplicate crate and add a bunch of shim implementations; we can also write generic code that works between Bitcoin and Elements...
* ...but we largely won't need to, because `bitcoin-consensus-encoding` ships with functions for decoding/encoding to `std::io` types, unbuffered `std::io` types, Vecs/slices, and hex strings, while `bitcoin-hashes` ships with functions for encoding into hash engines, and `bitcoin-io` with functions for decoding/encoding into `bitcoin::io` types.
* These traits use the "sans-IO paradigm", meaning that they can be used in no-std environments without the need for shim crates like `core2` (deleted from crates.io) or `bitcoin-io`, whose types tend to pollute every codebase that touches them.
* The traits also require no allocators: it is easy to encode even complex objects like `Transaction` without needing to copy any data into auxliary buffers, and some for decoding. If an object does not itself hold a `Box` or `Vec` it should be possible to encode/decode it completely allocator-free. Again, without shim crates like `ArrayVec` or whatever. Coupled with allocator-free encoders like `rust-bech32`, it is now possible to write arbitrary error correcting codes and encode bitcoin/elements types into them without needing any auxiliary buffers
* Because these traits are defined by a stable 1.0 crate, after *years* of the rust-bitcoin ecosystem fighting over API breakage and promising not to 2.0 basic stuff, they should remain stable across rust-elements versions. Since rust-elements is far from stable, this should make upgrades *much* easier since it will be increasingly possible to interoperate multiple versions of it.
* It paves the way for us to replace `Sequence` and `LockTime` with `bitcoin-units 1.0` types; this may invite type confusion between "bitcoin locktimes" and "elements locktimes" but in exchange we get stable types that have had years of bugfixes and API improvements. (And locktimes in Bitcoin, and therefore Elements, are notoriously fickle and hand to use correctly. Every API improvement we shipped found bugs in rust-miniscript.)
* Along the way, I made many API improvements to rust-elements that aren't directly related to encoding, but without which it was quite difficult to implement. These increase the size of this PR but they're all changes that need to be made eventually.


apoelstra:
Also, I renamed a few files, which is contributing to the huge diffstat. It's a big PR but it's not as big as it looks :).

src/internal_macros.rs
```diff
@@ -527,3 +527,547 @@ macro_rules! hex_deserialize(
...
+///     >);
+///     const ERROR_DISPLAY = "error decoding transaction output witness";
+///
+///     impl Decode for TxOut {
+///         fn convert_inner(output) -> Result<_, TxOutDecoderErrorInner> {
+///             let (asset, value, nonce, script_pubkey) = output;
+///             Ok(TxOut { asset, value, nonce, script_pubkey, witness: TxOutWitness::empty() })
+///         }
+///     }
+/// }
```
tcharding, line 919: Since you went to the trouble of writing such extensive docs is it worth adding dummy `Display` and `Error` impls here to remind folk?

src/transaction/pegin_witness.rs
```diff
@@ -0,0 +1,558 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Pegin Witnesses
+
+use core::fmt;
+use std::io;
+
+// lol should these be accessible from a more convenient path?
+use bitcoin::blockdata::block::BlockHashDecoder as GenesisHashDecoder;
```
tcharding: You could drop the `blockdata`. Do you still think the path is too nested?

tcharding:
FWIW this looks sane to me. I just tabbed through quickly to see how you did it.

delta1:
Looks good, will review ASAP. In the meantime a few small CI issues

src/transaction/pegin_witness.rs
apoelstra: Without `blockdata` I think it's okay. Really I'd like to write `bitcoin::BlockHashDecoder` but I'm unsure if that's a sane thing to do (reexporting every single encoder, decoder and decodererror at the root). It's what I did here in rust-elements and we can see how badly it pollutes the docs.

tcharding: Yeah I don't think its worth the noise.

src/internal_macros.rs
apoelstra: In my experience the compiler always complains about these being missing. I think it's probably not worth making the docs even bigger.

apoelstra:
On 5eb4f8db16a0c1b17426d3b6fd06534b13341d65 successfully ran local tests

delta1 (approved):
ACK 5eb4f8db16a0c1b17426d3b6fd06534b13341d65; tested locally

PR #293, opened 2026-07-31
pset: mask pegin/issuance flags out of the input outpoint

### LINKED ISSUE #292: PSET extraction leaves pegin flag in TxIn.previous_output.vout
In elements-rust, `PartiallySignedTransaction::extract_tx()` copies the flagged PSET previous_output_index directly into `TxIn.previous_output.vout`, while also setting `TxIn::is_pegin`.

Elements Core removes the pegin/issuance flag bits from the in-memory outpoint and represents them separately. Consequently, `SighashCache` hashes the flagged outpoint, producing a different SegWit v0 sighash than Elements Core. Signatures created from an extracted pegin PSET are therefore rejected by Elements.

`extract_tx()` should mask the pegin and issuance bits from previous_output_index when constructing `TxIn.previous_output`, while retaining `is_pegin` and issuance state separately.

delta1:
The pegin and issuance flags are stored in the high bits of the PSET `previous_output_index`, but they are not part of the outpoint itself: Elements Core keeps them separate in memory and only re-adds them when serializing. `Input::issuance_ids` was computing the asset entropy over the flagged index, so any new issuance round-tripped through a PSET derived different asset and token IDs than Elements does. 

`extract_tx` already masked the flags, but did so inline. Add `Input::previous_outpoint`, which strips them (leaving the all-1's coinbase index alone, as it carries no flags), and use it from both places so the two cannot drift apart again. Also replace the bare `1 << 30` / `1 << 31` literals with named constants.

Fixes #292

### DIFF
--- a/src/pset/map/input.rs
+++ b/src/pset/map/input.rs
@@ -169,6 +169,15 @@ const PSBT_ELEMENTS_IN_ASSET_PROOF: u8 = 0x14;
 /// Note that this does not indicate actual blinding status,
 /// but rather the expected blinding status prior to signing.
 const PSBT_ELEMENTS_IN_BLINDED_ISSUANCE: u8 = 0x15;
+
+/// Bit of [`Input::previous_output_index`] indicating that the input is a pegin.
+const OUTPOINT_PEGIN_FLAG: u32 = 1 << 30;
+/// Bit of [`Input::previous_output_index`] indicating that the input has an issuance.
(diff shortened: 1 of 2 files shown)


apoelstra:
Nice. Principled fix.

apoelstra (approved):
ACK e30b0206dc691fcdc592f423521e91f94022ebf6; successfully ran local tests

PR #298, opened 2026-09-14
blind: verify rangeproofs for confidential issuance amounts

delta1:
`verify_tx_amt_proofs` pushed confidential issuance amount and inflation
keys commitments into the balance sum without verifying any rangeproof.
The commitment entered the aggregate sum as an unconstrained curve point,
making the value-conservation check satisfiable for an arbitrary
transaction carrying an issuance: a relying verifier could be induced to
accept a transaction that mints value out of nothing.

This PR verifies the accompanying rangeproof against the unblinded
issuance generator, mirroring Elements Core's `VerifyIssuanceAmount`:
the proof message is empty for issuances, and the generator is the
unblinded generator of the issued asset (resp. reissuance token). A
confidential issuance amount without a rangeproof is now rejected with
`VerificationError::RangeProofMissing`.


apoelstra:
Since this looks alarming -- note that the invalid issuance here is rejected by all nodes on the network, even by `testproposedblock`, and the bad data is covered by signature hashes (so if a wallet is fooled into signing this the signature will be harmlessly invalid).

So it's ultimately no more harmless than any other form of invalid transaction that you can construct and get through this library, which does not claim to be (and never will claim to be) a complete consensus implementation.

src/blind.rs
```diff
@@ -1102,20 +1102,43 @@ impl Transaction {
...
+                        token_id,
+                        &inp.witness.inflation_keys_rangeproof,
+                    ),
                 ];
-                for (amt, asset) in &arr {
+                for (amt, asset, rangeproof) in &arr {
+                    // Issuance pseudo-inputs are never asset-blinded: the
+                    // generator is the unblinded generator of the issued
+                    // (or reissuance token) asset.
+                    let gen = Generator::new_unblinded(secp, asset.into_tag());
```
apoelstra, line 1120: In 173321e6330336afd3d8386a676f4b13e3fec263:

Nit: moving this line was not necessary (and it causes an unnecessary generator computation in the `Value::Null` case).

apoelstra:
Because the doccomment for `verify_tx_amt_proofs` claims that it does this check, and it didn't, we should backport to 0.25.x onward.

apoelstra (approved):
ACK 173321e6330336afd3d8386a676f4b13e3fec263; successfully ran local tests

